PR #519, opened 2026-06-24
feat(consent): collect sms consent on the /consent re-consent screen

leahpeker:
## Overview

Follow-up to #486. On staging, the **guidelines** consent modal (`/consent`) shows correctly, but it only ever collected **guidelines** consent — so users routed there (already-onboarded users whose `guidelines_consent_at` is null) had no way to supply **sms** consent, leaving `sms_consent_at` permanently null. #486 added the sms checkbox to **onboarding**, but onboarding has already passed for these legacy re-consent users — closing the gap #486 flagged as a known follow-up.

Now `/consent` shows a **second checkbox for sms** when the user still needs it (`needsSmsConsent`), and **continue requires both** checkboxes. When sms consent isn't needed, the screen is unchanged (guidelines-only).

## Changes

- **backend**: new `AcceptConsentIn { accept_sms }` schema; `/accept-guidelines/` takes an optional `accept_sms` flag and stamps `sms_consent_at` when set **and still missing** — an existing timestamp is never overwritten. Default empty body keeps it backward-compatible.
- **frontend**: conditional sms checkbox on `ConsentScreen` (shown only when `needsSmsConsent`), both-required gating, `acceptGuidelines(acceptSms)` threaded through the store + api client.
- regenerated `openapi_schema.json` + `types.gen.ts`.

**Design decisions (per discussion):**
- sms checkbox is **conditional** — only renders when `needsSmsConsent` is true.
- on `/consent`, sms is **required** (both must be checked to continue), unlike onboarding/the gate where sms is recorded-not-gated. The guidelines hard-gate is unchanged.
- existing consents are never overwritten (stamping guards on `is None`).

## Test plan

- [ ] Backend: `make agent-ci` (new sms cases in `test_accept_guidelines.py`)
- [ ] Frontend: `make agent-frontend-lint && agent-frontend-typecheck && agent-frontend-test` (ConsentScreen both-required + passthrough cases)
- [ ] Staging: log in as an already-onboarded user with `guidelines_consent_at` AND `sms_consent_at` null → `/consent` shows **both** checkboxes; continue disabled until both checked; lands on `/calendar`; `sms_consent_at` is now set
- [ ] Staging: a user needing only guidelines (sms already set) sees **just** the guidelines checkbox


frontend/src/api/auth.ts
```diff
@@ -153,10 +153,14 @@ export async function completeOnboarding(payload: {
   return mapUser(data);
 }
 
-export async function acceptGuidelines(): Promise<User> {
+export async function acceptGuidelines(acceptSms = false): Promise<User> {
   // Clears the hard guidelines-consent gate. Returns the updated user so the
-  // store can drop needsGuidelinesConsent and let the gate release.
-  const { data } = await apiClient.post<WireUser>('/api/auth/accept-guidelines/');
+  // store can drop needsGuidelinesConsent and let the gate release. When
```
leahpeker: his doesnt make sense to me - they should only hit this if acceptSms is true, right, so isnt this unnecessary?

actually can we rethink this to make this modal reuseable for whatever other needs come up. like can we just have it be an onNextLoginModal or soemthing along those lines and then we can create new needs as they arise. so like for example, all users must have consented to the guidelines and must have consented to the sms policy. those are really 2 standalone things. it's possible that in the future we have a diff thing, like maybe email policy. first we should have a check for any of the consents being missing, then condiitonally format the checkbox lines, and hten submit with all reqired

endpoint should expect which types of consent it's receiving and then store that -- so we can add to that be endpt in the future too

frontend/src/screens/auth/ConsentScreen.tsx
```diff
@@ -11,20 +11,29 @@ import { extractApiError } from '@/utils/errors';
 // landing page. Reached via OnboardingGate redirecting consent-pending users
```
leahpeker: there are way too many verbose comments going on, please reduce it down to 1 line max

frontend/src/screens/auth/ConsentScreen.tsx
```diff
@@ -11,20 +11,29 @@ import { extractApiError } from '@/utils/errors';
...
+  const blocked = !consented || (needsSms && !smsConsented);
+
   async function onAccept() {
-    if (!consented) return;
+    if (blocked) return;
     setServerError(null);
     setSubmitting(true);
     try {
-      await acceptGuidelines();
+      await acceptGuidelines(needsSms ? smsConsented : false);
```
leahpeker: i think we should have await acceptGuidlines and await acceptSmsConsent as 2 diff things rather than entangling this function

frontend/src/api/auth.ts
leahpeker: this has been addressed by the registry refactor in the later commits — there's now a consent registry (`frontend/src/models/consent.ts` mirroring `backend/users/_consents.py`) that's the single source of truth. `missingConsents(user)` checks for *any* outstanding consent, the checkbox lines are formatted from registry descriptors (no per-type branching), and submit sends all required types via `acceptConsents(consentTypes[])`. the `/accept-consents/` endpoint accepts `consent_types` and stores whichever it receives, so adding a future consent (e.g. email policy) is just one registry entry on each end. the conditional you flagged at this line is gone.

frontend/src/screens/auth/ConsentScreen.tsx
leahpeker: done — collapsed the header comment block down to a single line.

frontend/src/screens/auth/ConsentScreen.tsx
leahpeker: flagging a conflict here before i change this. this `onAccept` deliberately makes one batched `acceptConsents(acceptedTypes)` call because the registry refactor (per your earlier comment on auth.ts asking to "submit with all required" and make the endpoint accept which consent types it receives) was built specifically to avoid per-type branching and scale to future consents (e.g. email policy) with zero new code here. splitting into separate `await acceptGuidelines()` / `await acceptSmsConsent()` calls would re-introduce the per-type branching the registry removed and make two round-trips for one user action.\n\ni didn't want to guess against your own earlier direction, so i've left this as-is for now and raised it as a design question. happy to split it the moment you confirm which way you want — the registry batch, or explicit per-type calls.

leahpeker:
merge conflicts with `main` need manual resolution. auto-resolution was skipped to avoid a risky merge.

conflicted files (all logic, not mechanical):
- `frontend/src/screens/auth/ConsentScreen.tsx`
- `frontend/src/screens/auth/OnboardingScreen.tsx`
- `frontend/src/api/auth.ts`

root cause: both this branch and `main` independently contain a "registry-driven consent collection" refactor (`useConsentChecklist` / `acceptConsents` on main vs the manual two-checkbox sms-consent flow on this branch), and the two implementations diverged. reconciling the sms-consent feature with main's new registry pattern is a design decision, not a mechanical merge — needs a human to resolve. branch was left untouched (rebase aborted).

backend/users/_consents.py
```diff
@@ -0,0 +1,49 @@
+"""Consent type registry — the single source of truth for the kinds of consent
```
leahpeker: this comment is far far too verbose

backend/users/models.py
```diff
@@ -56,7 +56,7 @@ class User(AbstractUser):
     # PERSISTENT USER STATE: when this user last accepted the community guidelines
```
leahpeker: let's get rid of all of these comment lines on the models please

backend/users/schemas.py
```diff
@@ -220,18 +221,21 @@ class ErrorOut(BaseModel):
     detail: str
 
 
+class AcceptConsentsIn(BaseModel):
+    # Which consents the user is accepting. Stamping is registry-driven and never
```
leahpeker: limit comment to 1 line

backend/users/schemas.py
```diff
@@ -220,18 +221,21 @@ class ErrorOut(BaseModel):
...
     # OptionalEmail (like the other schemas) so an empty-string email from the
     # client coerces to None instead of failing EmailStr validation.
     email: OptionalEmail = None
     # Inline consent collected during onboarding for users with no prior consent
-    # (admin-created accounts have no JoinRequest). Each flag stamps the matching
-    # *_consent_at only when True AND the user hasn't already consented — so an
-    # existing timestamp is never overwritten. Omitted/False = leave as-is.
-    accept_guidelines: bool = False
-    accept_sms: bool = False
+    # (admin-created accounts have no JoinRequest). Registry-driven; see _consents.
```
leahpeker: remove comment

backend/users/schemas.py
```diff
@@ -220,18 +221,21 @@ class ErrorOut(BaseModel):
...
+class AcceptConsentsIn(BaseModel):
+    # Which consents the user is accepting. Stamping is registry-driven and never
+    # overwrites an existing timestamp; see users._consents.
+    consent_types: list[ConsentType] = Field(default_factory=list)
+
+
 class OnboardingIn(BaseModel):
     new_password: str = Field(max_length=FieldLimit.PASSWORD)
     display_name: str | None = Field(default=None, max_length=FieldLimit.DISPLAY_NAME)
     # OptionalEmail (like the other schemas) so an empty-string email from the
```
leahpeker: remove comment here too

frontend/src/api/types.gen.ts
```diff
@@ -14,14 +14,16 @@ export interface paths {
...
         /**
-         * Accept Guidelines
-         * @description Stamp the current user's guidelines consent, clearing the hard gate.
+         * Accept Consents
+         * @description Record the consents the caller is accepting, clearing the relevant gates.
          *
-         *     Idempotent: re-accepting just re-stamps the timestamp. The gate (see
-         *     config.auth.GatedJWTAuth) treats a null guidelines_consent_at as "must
-         *     consent", so any non-null value satisfies it.
+         *     Registry-driven (see users._consents): each ConsentType in the payload stamps
```
leahpeker: overly verbose, let's reduce this to justl ike 1-2 lines

frontend/src/api/types.gen.ts
```diff
@@ -1586,6 +1588,11 @@ export interface paths {
 export type webhooks = Record<string, never>;
 export interface components {
     schemas: {
+        /** AcceptConsentsIn */
```
leahpeker, line 1567: remove comment

frontend/src/api/types.gen.ts
```diff
@@ -1586,6 +1588,11 @@ export interface paths {
...
+        AcceptConsentsIn: {
+            /** Consent Types */
```
leahpeker, line 1569: remove comment

frontend/src/api/types.gen.ts
```diff
@@ -1690,6 +1697,11 @@ export interface components {
             /** Reacted By Me */
             reacted_by_me: boolean;
         };
+        /**
```
leahpeker, line 1680: remove comment

frontend/src/api/types.gen.ts
```diff
@@ -2906,16 +2918,8 @@ export interface components {
...
-             * Accept Guidelines
-             * @default false
-             */
-            accept_guidelines: boolean;
-            /**
-             * Accept Sms
-             * @default false
-             */
-            accept_sms: boolean;
+            /** Consent Types */
```
leahpeker, line 2923: remove comment

frontend/src/models/consent.ts
```diff
@@ -0,0 +1,51 @@
+// Consent registry — the single source of truth for the kinds of consent the
```
leahpeker: remove comment here

frontend/src/models/consent.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const ConsentType = {
+  Guidelines: 'guidelines',
+  Sms: 'sms',
+} as const;
+export type ConsentTypeValue = (typeof ConsentType)[keyof typeof ConsentType];
+
+export interface ConsentDescriptor {
+  type: ConsentTypeValue;
+  // Whether this consent is still outstanding for the given user.
```
leahpeker: remve comments

frontend/src/models/consent.ts
```diff
@@ -0,0 +1,51 @@
...
+  needs: (user: User) => boolean;
```
leahpeker: name should be clearer - let's try like hasConsented or something

frontend/src/screens/auth/ConsentChecklist.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface ConsentChecklistProps {
+  // The consents to render a checkbox for — typically missingConsents(user).
+  consents: readonly ConsentDescriptor[];
+  // Currently-checked consent types (controlled by the parent).
+  checked: ReadonlySet<ConsentTypeValue>;
+  onToggle: (type: ConsentTypeValue, checked: boolean) => void;
+}
+
+// Renders one checkbox per outstanding consent, driven by the consent registry.
```
leahpeker: dont need this comment either

frontend/src/screens/auth/ConsentChecklist.tsx
```diff
@@ -0,0 +1,42 @@
+import { Link } from 'react-router-dom';
+import type { ConsentDescriptor, ConsentTypeValue } from '@/models/consent';
+
+interface ConsentChecklistProps {
+  // The consents to render a checkbox for — typically missingConsents(user).
```
leahpeker: get rid of the comments

frontend/src/screens/auth/ConsentScreen.test.tsx
```diff
@@ -57,20 +68,60 @@ describe('ConsentScreen', () => {
...
+  it('requires both checkboxes when sms consent is also needed', async () => {
+    mockStore({ needsSmsConsent: true });
+    render(
+      <MemoryRouter>
+        <ConsentScreen />
+      </MemoryRouter>,
+    );
+    const continueBtn = screen.getByRole('button', { name: /continue/i });
+    expect(continueBtn).toBeDisabled();
+    // Ticking only guidelines is not enough — sms is still required.
```
leahpeker: remove comment

backend/tests/test_accept_guidelines.py
```diff
@@ -1,7 +1,8 @@
-"""Tests for POST /api/auth/accept-guidelines/ — the consent-gate escape hatch.
```
leahpeker, line 1: remove comment

PR #520, opened 2026-06-24
fix: 506 - graceful handling for malformed role permissions

### LINKED ISSUE #506: permissions bug
## Context

A user who has the "official pda event selection" permission but does not hold an admin role hit a crash when trying to add the official pda event type. Reported via community feedback. There are two distinct problems:

1. The flow crashes instead of degrading gracefully.
2. This permission alone should allow adding the official pda event type

## What needs to happen

- Make the failure path graceful: surface a clear error instead of crashing.
- Audit the permission check for adding the official pda event type to confirm the correct permission(s) gate it, and that a user holding "official pda event selection" without admin behaves as intended.

## Acceptance criteria

- [ ] attempting to add the official pda event type without sufficient permission no longer crashes; the user sees a clear, graceful error
- [ ] the permission required to add the official pda event type is verified to be correct and enforced consistently on frontend and backend
- [ ] a user with "official pda event selection" permission (no admin role) gets the intended behavior (either allowed, or cleanly denied)

## Open Questions

- Exact crash location / stack trace is unknown.
- Whether "official pda event selection" alone is meant to permit adding the official pda event type needs a product/permissions decision.

---

## Original

user with official pda event selection permissions but not admin roles hit a crash when trying to add the official pda event type

twofold bug 
1. it shouldnt crash, there should be graceful errors
2. potential permissions issue

## Metadata

- **Route:** `/profile`

leahpeker:
## Overview

A user with the `tag_official_event` permission but no admin role hit a render-time crash when adding the official pda event type (reported via community feedback). `hasPermission` called `.includes` on a role's `permissions`, which rides the wire from a backend `JSONField` with no DB-level shape guarantee — a non-array value (legacy/corrupt row) threw during render, and the route error boundary then surfaced the raw JS error string to the user instead of a friendly message.

This makes the failure path graceful and — the core fix — **pushes the `permissions: string[]` guarantee down to the backend** so no consumer has to defend against malformed data on the wire:

- **Backend is now the single source of the invariant.** `Role.effective_permissions` coerces the underlying `JSONField` to a clean `list[str]` (non-list → `[]`, non-string entries filtered), so a corrupt/out-of-band row can't reach any client as null/object/non-array. `RoleOut.permissions` is therefore guaranteed for every API consumer. `User.has_permission` routes through `effective_permissions` too, replacing `key in (role.permissions or [])`, which silently mis-matched on a `str`/`dict` row.
- **Frontend `normalizePermissions()` stays as defense-in-depth**, not as the guarantee — applied at the role-mapping boundaries (`api/auth.ts`, `api/users.ts`, `api/roles.ts`) and reused by `hasPermission`, so render can't throw even if the backend invariant ever regresses.
- `RootRouteError` now shows a friendly fallback (`something went wrong — try again or head back home`) for unexpected JS exceptions instead of leaking the raw `error.message`; the technical detail still goes to `reportError`, and route-error responses (404s) keep their status line.
- Regression tests on both sides: backend tests for the reachable corrupt shapes (`str`/`dict`/`int` → no perms) and non-string list entries (`null` is unreachable — the column is `NOT NULL`); frontend tests for the non-array guard and the helper.

**Permission audit:** `tag_official_event` (BE `PermissionKey.TAG_OFFICIAL_EVENT` / FE `Permission.TagOfficialEvent`) correctly and consistently gates official events — the frontend filters the "official" visibility option and the backend returns a clean 403 on `POST /events/` when `event_type=official` without the permission. A user holding the permission without an admin role is **intentionally allowed** to create official events; no permission-model change was needed.

Fixes #506

## Test plan

- [x] backend: `TestRolesAndPermissions` (incl. corrupt-shape regressions) — pass
- [x] backend: `make agent-lint` + `make agent-typecheck` — pass
- [x] frontend: `make agent-frontend-test` (incl. `permissions.test.ts`) — pass
- [x] frontend: `make agent-frontend-lint` + format-check — pass
- [ ] manual: as a user with `tag_official_event` and no admin role, open the event form and add an official event — no crash, option available
- [ ] manual: with a deliberately malformed role (non-array `permissions`), the app degrades to "no perms" instead of white-screening

### DIFF
--- a/backend/users/roles.py
+++ b/backend/users/roles.py
@@ -29,16 +29,15 @@ class Role(models.Model):
 
     @property
     def effective_permissions(self) -> list[str]:
-        """Admin role implicitly grants every permission (see User.has_permission).
-        Return the current PermissionKey set so the UI reflects reality even if
-        the DB row was seeded before newer keys were added.
+        """The permission keys this role grants (see User.has_permission).
 
-        Coerces the JSONField to a clean list[str] (corrupt/out-of-band rows may
-        hold non-list/non-string values) so RoleOut.permissions holds for every consumer.
+        The default admin role implicitly grants every current key. Other roles
+        return their stored keys, coerced to a clean list[str] — the JSONField
+        defaults to a list, but legacy/corrupt rows may hold other values.
         """
         if self.name == "admin" and self.is_default:
             return list(PermissionKey.values)
-        raw = self.permissions
-        if not isinstance(raw, list):
+        stored = self.permissions
+        if not isinstance(stored, list):
             return []
-        return [p for p in raw if isinstance(p, str)]
+        return [p for p in stored if isinstance(p, str)]
--- a/backend/users/roles.py
+++ b/backend/users/roles.py
@@ -32,7 +32,13 @@ class Role(models.Model):
         """Admin role implicitly grants every permission (see User.has_permission).
         Return the current PermissionKey set so the UI reflects reality even if
         the DB row was seeded before newer keys were added.
+
+        Coerces the JSONField to a clean list[str] (corrupt/out-of-band rows may
+        hold non-list/non-string values) so RoleOut.permissions holds for every consumer.
         """
         if self.name == "admin" and self.is_default:
             return list(PermissionKey.values)
-        return list(self.permissions)
+        raw = self.permissions
+        if not isinstance(raw, list):
+            return []
+        return [p for p in raw if isinstance(p, str)]
--- a/backend/users/roles.py
+++ b/backend/users/roles.py
@@ -29,12 +29,7 @@ class Role(models.Model):
 
     @property
     def effective_permissions(self) -> list[str]:
-        """The permission keys this role grants (see User.has_permission).
-
-        The default admin role implicitly grants every current key. Other roles
-        return their stored keys, coerced to a clean list[str] — the JSONField
-        defaults to a list, but legacy/corrupt rows may hold other values.
-        """
+        """Permission keys this role grants: all keys for the default admin role, else the stored keys coerced to a clean list[str] (legacy/corrupt rows may hold non-list or non-string values)."""
         if self.name == "admin" and self.is_default:
             return list(PermissionKey.values)
         stored = self.permissions
--- a/frontend/src/models/permissions.ts
+++ b/frontend/src/models/permissions.ts
@@ -42,10 +42,9 @@ export interface UserLike {
 // Backend sends permissions: string[] (Role.effective_permissions coerces the
 // JSONField). Defense-in-depth so a regression can't make render throw; the
 // null/undefined cases cover a field the API dropped entirely.
-export function normalizePermissions(
-  value: readonly string[] | null | undefined,
-): string[] {
-  return Array.isArray(value) ? [...value] : [];
+export function normalizePermissions(value: readonly string[] | null | undefined): string[] {
+  if (!Array.isArray(value)) return [];
+  return value.filter((v): v is string => typeof v === 'string');
 }
 
 export function hasPermission(user: UserLike | null, key: PermissionKey): boolean {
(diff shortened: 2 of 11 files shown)


backend/users/roles.py
```diff
@@ -30,9 +30,15 @@ def effective_permissions(self) -> list[str]:
         """Admin role implicitly grants every permission (see User.has_permission).
         Return the current PermissionKey set so the UI reflects reality even if
         the DB row was seeded before newer keys were added.
+
+        Coerces the JSONField to a clean list[str] (corrupt/out-of-band rows may
+        hold non-list/non-string values) so RoleOut.permissions holds for every consumer.
         """
         if self.name == "admin" and self.is_default:
             from users.permissions import PermissionKey
```
leahpeker: this import should be on the top of the file

backend/users/roles.py
```diff
@@ -30,9 +30,15 @@ def effective_permissions(self) -> list[str]:
...
 
             return list(PermissionKey.values)
-        return list(self.permissions)
+        raw = self.permissions
+        if not isinstance(raw, list):
```
leahpeker: how could it not be a list? can it be null? also rename raw to be more useful pls

backend/users/roles.py
```diff
@@ -30,9 +30,15 @@ def effective_permissions(self) -> list[str]:
         """Admin role implicitly grants every permission (see User.has_permission).
```
leahpeker: overly verbose

frontend/src/models/permissions.ts
```diff
@@ -41,10 +41,20 @@ export interface UserLike {
...
 
+// Backend guarantees permissions: string[] (Role.effective_permissions coerces
+// the JSONField). Kept as defense-in-depth so render can't throw if that regresses.
+export function normalizePermissions(value: unknown): string[] {
+  return Array.isArray(value) ? (value as string[]) : [];
+}
+
 export function hasPermission(user: UserLike | null, key: PermissionKey): boolean {
   if (!user) return false;
+  // Re-normalize per role: a UserLike can be built in tests or from a source
```
leahpeker: overly verbose

frontend/src/models/permissions.ts
```diff
@@ -41,10 +41,20 @@ export interface UserLike {
   }[];
 }
 
+// Backend guarantees permissions: string[] (Role.effective_permissions coerces
+// the JSONField). Kept as defense-in-depth so render can't throw if that regresses.
+export function normalizePermissions(value: unknown): string[] {
```
leahpeker: can we modify this so type IS known? we get this from the BE so we should be able to infer type

backend/users/roles.py
leahpeker: done — hoisted `from users.permissions import PermissionKey` to the top of the file (no circular import; permissions.py only imports django.db).

backend/users/roles.py
leahpeker: the JSONField defaults to a list, but legacy/corrupt/out-of-band rows can hold anything (incl. null) — that's the bug this PR guards against — so the isinstance check stays. renamed `raw` -> `stored` and tightened the docstring to say so.

backend/users/roles.py
leahpeker: trimmed the docstring down.

frontend/src/models/permissions.ts
leahpeker: trimmed the comments.

frontend/src/models/permissions.ts
leahpeker: typed the input as `readonly string[] | null | undefined` — the real BE contract (RoleOut.permissions is list[str], and null/undefined cover a dropped field) — instead of `unknown`. kept the runtime guard as defense-in-depth since it's exactly the regression this PR fixes.

PR #522, opened 2026-06-24
fix(photos): drop unsupported heic uploads and surface crop decode errors (Issue 505)

### LINKED ISSUE #505: photo upload didnt work
## Context

A user reported that event photo upload did not work. This is a thin bug report from community feedback — it lacks any reproduction details, so the failure mode is currently unknown.

## What needs to happen

Reproduce the failure, identify whether the break is on the frontend (upload control, request) or backend (storage/proxy, validation), and fix it so event photo upload works reliably.

## Acceptance criteria

- [ ] the failure is reproduced and root-caused
- [ ] a user can successfully upload an event photo and see it persist
- [ ] if the upload fails, the user sees a clear error rather than a silent failure

## Open Questions

- No reproduction steps provided.
- No error message or browser console state captured.
- Unknown which photo / which event / which user role was involved.
- Needs reproduction before a fix can be scoped.

---

## Original

event photo upload didnt work for a user

## Metadata

- **Route:** `/profile`

leahpeker:
## Summary

Closes #505

A user reported "event photo upload didn't work" from `/profile`. Root cause: the app **advertised heic/heif support it can't deliver**. The client-side crop pipeline (`ImageCropDialog` → `cropImage`) decodes images via a browser ``/canvas, which **cannot decode heic/heif** — the default format for Apple/macOS photos. A macOS Chrome user (matching the issue's user agent) picked a `.heic` photo → it passed validation → the crop dialog's `` never fired `onLoad` → the "save" button stayed permanently disabled with **no error**. A silent failure.

- **Drop heic/heif** from the frontend `accept`/validation (`AvatarUpload.tsx`, `EventFormPhoto.tsx`) and the backend allowed-type sets (`users/_auth.py`, `community/_event_schemas.py`), so the upload contract matches what the crop pipeline can actually process. Updated the user-facing format copy.
- **Surface decode failures**: added an `onError` handler + `role="alert"` message to `ImageCropDialog` so any image that fails to decode shows a clear error instead of a dead "save" button (satisfies the "clear error rather than silent failure" acceptance criterion).
- **Tests**: backend rejects heic on both the profile and event endpoints; a frontend test fires the img `error` event and asserts the alert appears and save stays disabled.

## Test plan

- [ ] backend: `cd backend && uv run pytest tests/test_photos.py --create-db -n 0` — 22 passed (full suite 1005 passed on a fresh DB)
- [ ] frontend: `cd frontend && pnpm test src/components/ImageCropDialog.test.tsx` — 5 passed; `pnpm lint`, `pnpm format:check`, `pnpm typecheck` clean
- [ ] manual: on a Mac, open `/profile` and `/settings`, pick a `.heic` photo → it's no longer offered in the file picker; pick a jpeg/png/webp/gif → crop + save persists
- [ ] manual: if an image fails to decode in the crop dialog, a clear error appears instead of a stuck "save" button

> Note: the local test DB carried an `is_member NOT NULL` column from another branch (Issue 491, not in this branch); run backend tests with `--create-db`. Parallel-load timeouts in the auth/join screen tests are pre-existing and unrelated to this diff (those files don't import anything changed here and pass in isolation).

### DIFF
--- a/backend/community/_event_schemas.py
+++ b/backend/community/_event_schemas.py
@@ -404,11 +404,12 @@ class EventPatchIn(BaseModel):
 
 
 _MAX_EVENT_PHOTO_SIZE = 10 * 1024 * 1024  # 10 MB
+# heic/heif are intentionally excluded: the frontend crops via a browser canvas
+# that can't decode them, so such uploads can only arrive via a hand-crafted
+# request and would never round-trip through the real UI (issue 505).
 _ALLOWED_IMAGE_TYPES = {
     "image/jpeg",
     "image/png",
     "image/webp",
     "image/gif",
-    "image/heic",
-    "image/heif",
 }
--- a/backend/tests/test_photos.py
+++ b/backend/tests/test_photos.py
@@ -100,6 +100,12 @@ class TestProfilePhoto:
         response = api_client.post("/api/auth/me/photo/", {"photo": big}, **_auth(member))
         assert response.status_code == 400
 
+    def test_upload_rejects_heic(self, api_client, member):
+        # heic can't be cropped client-side, so we don't accept it (issue 505).
+        heic = SimpleUploadedFile("photo.heic", b"fake heic bytes", content_type="image/heic")
+        response = api_client.post("/api/auth/me/photo/", {"photo": heic}, **_auth(member))
+        assert response.status_code == 400
+
     def test_delete_photo(self, api_client, member):
         photo = _make_test_image()
         api_client.post("/api/auth/me/photo/", {"photo": photo}, **_auth(member))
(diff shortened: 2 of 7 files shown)


leahpeker:
the premise of this work is wrong - we SHOULD support HEIC and HEIF uploads but the inclination to catch the error to show an error instead of a stuck Save button is on point

leahpeker:
Closing as no longer needed. The original bug (Issue 505) was a silent failure on event photo upload, which was root-caused and fixed by two subsequent PRs: #528 (seed crop on image load so save fires the upload) and #535 (default rect crop to full photo). heic/heif support is working and should remain as-is. The premise of this PR — that heic was unsupported and should be dropped — was incorrect.

Issue #403, opened 2026-04-28
event text blasts


leahpeker:

leahpeker:
closing — this was scoped as a server-side SMS text blast via twilio, which requires a sendable phone number we can't obtain. the twilio/sms backend started on `feat/text-blasts-403` (PR #414, draft) is dormant per the broader sms→email pivot (#432).

replacing with two new issues that work within our actual constraints:
- **email blast to attendees** (server-side, via resend — already wired up)
- **group text to attendees** (client-side `sms:` link / group-text, no server number needed)

PR #524, opened 2026-06-24
feat(events): 500 - hosts can start a group text to attendees

### LINKED ISSUE #500: feat(events): hosts can start a group text to event attendees
Give event hosts/co-hosts a button that opens their phone's Messages app pre-populated with all attendee phone numbers — effectively starting a **group text**. We can't do a true server-side text blast (no sendable phone number — see closed #403), but the host's *own* phone can send a group message for free. This is the client-side counterpart to the email blast (#499).

> Supersedes/implements the idea in #103. Close #103 as a duplicate of this once this lands (don't close it yet).

## Approach
No server send. The host's device sends the message. Use the `sms:` URI scheme with comma-separated recipients:
- iOS: `sms:+1234567890,+0987654321` (multiple recipients → group thread)
- Android: similar `sms:` intent support
- Desktop/web (no `sms:` handler): **fall back to a "copy all numbers" button** so the host can paste into whatever they use.

## Scope (frontend-only — likely no backend)

- **Entry point:** event detail page, host/co-host only — same kebab/3-dot menu as the email blast (#499). Gate on the same host/co-host check the client already uses for host-only controls.
- **Recipients:** attendee phone numbers. The host already has access to guest phone numbers today (they're shown in the attendee list to hosts). Pull the numbers the client already has for the attendee list — **confirm whether the event payload exposes attendee phone numbers to the host client; if not, that's the one backend touch needed** (a host-gated endpoint returning attendee phones for this event). **Default audience is everyone who RSVP'd — including `can't go`** (they still want updates: time/location changes, cancellations), matching #499. Optionally let the host narrow by status.
- **Behavior:**
  - On tap → build `sms:<comma-joined E.164 numbers>` and open it (anchor href / `window.location`).
  - Skip attendees with no phone; if some are skipped, note it ("N attendees have no number and won't be included").
  - **Desktop fallback:** if `sms:` isn't supported, show a "copy all numbers" action that copies the comma-separated list to clipboard, with a toast.
- **All copy lowercase** (CLAUDE.md frontend casing rule).

## Out of scope
- Server-side SMS / Twilio — never.
- Message templating / pre-filled body text (could add `?body=` later; v1 just opens the thread with recipients).
- Email blast — that's #499.

## Acceptance criteria
- Host/co-host sees the action; non-hosts don't.
- Tapping on a mobile device opens Messages with all eligible attendee numbers as a group thread.
- No-phone attendees are excluded and the count is surfaced.
- On desktop (no `sms:` handler), the copy-all-numbers fallback works.

## Tests
**Frontend:** action visible only to host/co-host; builds correct `sms:` URI with comma-joined E.164 numbers for the chosen audience; default audience includes `can't go` RSVPs; excludes no-phone attendees and reports the count; desktop fallback copies the list to clipboard and toasts; optional status-narrowing filter works when supplied.
(If a backend endpoint is needed to expose attendee phones: host-gated, non-host → 403/404, returns only this event's attendee numbers.)

## Verify
`make agent-ci` (or `make agent-frontend-ci`) before done.

leahpeker:
## Overview

Give event hosts/co-hosts a button that opens their phone's Messages app pre-populated with all attendee phone numbers — effectively starting a group text. There is no server-side text blast (no sendable number — see #403); the host's own device sends the message. On desktop (no `sms:` handler) it falls back to a "copy all numbers" action.

- New `utils/groupText` helpers: `collectRecipients` (dedupe E.164, count no-phone skips), `buildSmsUri` (comma-joined `sms:` group thread), `isSmsSupported` (mobile-UA heuristic — no real feature detection for the scheme).
- New `GroupTextButton` rendered in the rsvp card on the event detail page, gated on `isCoHost`. This mirrors the backend's existing host-only phone visibility (`_can_see_phones`, creator + co-hosts), so **no backend change is needed** — non-hosts already receive `phone: null` and the button stays disabled with no numbers.
- Default audience is everyone who RSVP'd, including "can't go" (they still want time/location/cancellation updates).
- No-phone attendees are excluded and the skipped count is surfaced on both the mobile `sms:` path (`toast.info`) and the desktop copy fallback (success toast).
- All copy is lowercase per the frontend casing rule.

Closes #500

## Test plan

- [x] unit: `collectRecipients` collects across all rsvp statuses incl. can't-go, dedupes, excludes/counts no-phone
- [x] unit: `buildSmsUri` comma-joins / returns null when empty; `isSmsSupported` true on mobile UA, false on desktop
- [x] component: mobile opens `sms:` group thread with all eligible numbers; desktop copies the list + toasts; skipped count surfaced on both paths; action disabled when no attendee has a number
- [x] `make agent-frontend-lint` / `typecheck` / `format-check` pass
- [ ] manual: on a phone, tap "text attendees" → Messages opens with attendees as a group thread
- [ ] manual: on desktop, "copy attendee numbers" copies the comma-separated list

> Note: the 4 `JoinScreen.test.tsx` failures seen under the default 5s parallel test run are a pre-existing timeout flake (unrelated files; pass in isolation and with a raised timeout). The full suite is green (559 passed / 1 skipped) with an adequate timeout.

## Screenshots

TODO


frontend/src/screens/events/GroupTextButton.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+    error: (m: string) => {
+      toastError(m);
+    },
+    info: (m: string) => {
+      toastInfo(m);
+    },
+  },
+}));
+
+const isSmsSupported = vi.fn<() => boolean>();
```
leahpeker: why do we want to check this? sms is supported by default and all users have phone numbers on the site

frontend/src/screens/events/GroupTextButton.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+vi.mock('@/utils/groupText', async (importOriginal) => {
+  const actual = await importOriginal<typeof GroupTextModule>();
+  return { ...actual, isSmsSupported: () => isSmsSupported() };
+});
+
+import { GroupTextButton } from './GroupTextButton';
+
+const BASE_EVENT: Event = {
```
leahpeker: dont we have test utils or something? if not, let's start making some bc this could be shared with other test files

frontend/src/utils/groupText.ts
```diff
@@ -0,0 +1,51 @@
+// Helpers for the host "group text" action — opens the host's own Messages app
```
leahpeker: let's avoid overly verbose comments pls, apply everywhere

frontend/src/screens/events/GroupTextButton.test.tsx
leahpeker: removed — that path (and the no-phone-skip test) covered scenarios that aren't real: sms is supported by default and all users have phone numbers. dropped them.

frontend/src/screens/events/GroupTextButton.test.tsx
leahpeker: done — added a shared makeEvent(overrides) factory in frontend/src/test/fixtures.ts and switched this test to it. other files can reuse it now; migrating the existing test files is tracked in #527.

frontend/src/utils/groupText.ts
leahpeker: trimmed the verbose comments here, keeping only the high-value 'why' notes; applied the same trimming across the files changed in this PR.

leahpeker: trimmed the verbose comments here and applied the lighter style across the new group-text files (groupText.ts + GroupTextButton.tsx).

frontend/src/screens/events/GroupTextButton.test.tsx
leahpeker: added shared test fixtures — `frontend/src/test/fixtures.ts` now exports a `makeEvent` factory, and this test (plus others) uses it instead of inlining event objects.

frontend/src/screens/events/GroupTextButton.test.tsx
leahpeker: this check is about the **host'\''s device**, not the recipients — `isSmsSupported()` sniffs the user agent to tell mobile from desktop. on mobile we open an `sms:` group thread; on desktop `sms:` links are dead, so we fall back to copying the numbers to the clipboard. so it'\''s not about whether members have phone numbers (they do).

happy to drop the desktop fallback and always open `sms:` if you'\''d prefer simpler — the tradeoff is a desktop host gets a no-op button. let me know which way you want it and i'\''ll adjust. leaving this thread open for your call.

frontend/src/utils/groupText.ts
```diff
@@ -0,0 +1,37 @@
+// Host "group text" helpers — the host's device sends (see #403).
```
leahpeker: remove comment

frontend/src/utils/groupText.ts
```diff
@@ -0,0 +1,37 @@
...
+      continue;
+    }
+    if (seen.has(phone)) continue;
+    seen.add(phone);
+    phones.push(phone);
+  }
+  return { phones, skippedCount };
+}
+
+// Comma-joined recipients open as one group thread on iOS and Android.
```
leahpeker: remove comment

frontend/src/utils/groupText.ts
```diff
@@ -0,0 +1,37 @@
...
+export function buildSmsUri(phones: string[]): string | null {
+  if (phones.length === 0) return null;
+  return `sms:${phones.join(',')}`;
+}
+
+// No feature detection for `sms:` handlers exists, so approximate via mobile UA.
```
leahpeker: remove comment

frontend/src/utils/groupText.ts
leahpeker: done — removed in edeef5f

frontend/src/utils/groupText.ts
leahpeker: done — removed in edeef5f

frontend/src/utils/groupText.ts
leahpeker: done — removed in edeef5f

frontend/src/screens/events/GroupTextButton.test.tsx
leahpeker: the check isn't about recipients having numbers — it's about the *host's own device*. `isSmsSupported()` approximates whether the host is on mobile (where the `sms:` link opens a group thread) vs desktop (where it does nothing). on desktop the button falls back to copying the numbers to the clipboard instead.

so the real question is: do we want to support hosts opening this on a laptop? if we assume hosts are always on mobile, i can drop `isSmsSupported`, the clipboard fallback, and the 'copy attendee numbers' label and simplify this down a lot. flagging that decision for you — happy to go either way.

frontend/src/api/textRecipients.ts
```diff
@@ -0,0 +1,53 @@
+// Host-only group-text recipients (see #500).
```
leahpeker: remove comment

frontend/src/screens/events/GroupTextButton.tsx
```diff
@@ -0,0 +1,58 @@
+// Host-only "group text" trigger (see #500). Opens a dialog that fetches the
```
leahpeker: remove comment

frontend/src/utils/groupText.ts
```diff
@@ -0,0 +1,69 @@
+import type { TextRecipients } from '@/api/textRecipients';
+
+// The recipient groups a host can pick when texting attendees. Each value is a
```
leahpeker: remove the comments in this file

PR #538, opened 2026-06-24
docs: 530 - explore is_admin concept for admin/permission checks

### LINKED ISSUE #530: Investigate an is_admin concept to simplify admin/permission checks
## Context

Raised in review of PR #521 (issue #491). Recipient/permission queries currently express "is this user an admin-or-event-manager" with a repeated two-part filter, e.g. in `backend/notifications/service.py:165`:

```python
UserModel.objects.members().filter(
    Q(roles__name="admin", roles__is_default=True)
    | Q(roles__permissions__contains=PermissionKey.MANAGE_EVENTS)
).distinct()
```

The reviewer asked: can we avoid repeating this `admin role OR permission` check everywhere? One idea floated was an `is_admin` flag that is **only** true for the role literally named `admin`, so call sites can do a clean `is_admin` check instead of the compound role/permission `Q` filter — possibly as part of a broader rethink of how roles/permissions are modeled.

This affects role architecture repo-wide (not just PR #521), so it's an investigation, not a directed change.

## What needs to happen

Investigate and recommend whether to introduce an `is_admin` concept (and/or otherwise simplify the admin/permission checks), covering:

- Enumerate every site that hand-rolls the `admin role OR <permission>` filter (grep for `roles__name="admin"`, `roles__permissions__contains`, `is_default=True`) and characterize what each is really asking.
- Evaluate options: (a) a derived `is_admin` property/manager method, (b) a stored `is_admin` field kept in sync with the admin role, (c) a permission-only check with no special-casing of the named admin role, (d) leave as-is. Weigh correctness, drift risk (stored flag vs role membership), and how it interacts with the existing `Role`/`PermissionKey` model.
- Recommend one approach with a migration/refactor sketch and the blast radius (which files change).

Output: a short written recommendation (spec/doc), not necessarily an implementation — the implementation can be a follow-up once the direction is agreed.

## Acceptance criteria

- [ ] All call sites using the compound `admin role OR permission` filter are enumerated
- [ ] Options for simplifying (incl. an `is_admin` flag) are evaluated with trade-offs
- [ ] A recommended approach is documented with blast radius and a migration/refactor sketch
- [ ] No production behavior change in this issue (investigation/spec only)

leahpeker:
## Overview

Investigation spike for issue #530 — should we introduce an `is_admin` concept to simplify the repeated `admin role OR <permission>` checks?

This PR adds a **findings spec only** (`docs/superpowers/specs/2026-06-24-is-admin-concept-explore.md`); it changes **no production code**. Key findings:

- The compound `Q(roles__name="admin", roles__is_default=True) | Q(roles__permissions__contains=...)` filter is duplicated in **exactly three places, all in `backend/notifications/service.py`** (join-request, event-flag, and magic-link recipient queries). Every other admin/permission site already goes through `User.has_permission()`, whose admin short-circuit keeps them clean.
- A single-user admin check (`_helpers._is_admin`) and an "admin grants every permission" rule (`Role.effective_permissions`, `User.has_permission`) already exist — only the *queryset* case is duplicated.
- The frontend already derives admin status from serialized roles; no `is_admin` field is needed on the wire.

**Recommendation:** add a derived `User.objects.with_permission(key)` queryset helper (no migration, no drift, preserves the `.distinct()` de-dup invariant) and replace the three notification filters with it. Reject a stored `is_admin` boolean — it adds a second source of truth and wouldn't even remove the permission half of the compound filter. Implementation is a follow-up.

Closes #530

## Test plan

- [ ] N/A — documentation/spec only, no code change. CI (`make agent-ci`) green on the docs change.

### DIFF
--- a/docs/superpowers/specs/2026-06-24-is-admin-concept-explore.md
+++ b/docs/superpowers/specs/2026-06-24-is-admin-concept-explore.md
@@ -1,0 +1,269 @@
+# Explore: Investigate an `is_admin` concept to simplify admin/permission checks (#530) — Findings
+
+**Date:** 2026-06-24
+**Issue:** https://github.com/ProteinDeficientsAnonymous/pda/issues/530
+**Branch / PR:** `auto-530-is-admin-explore` (draft PR linked from the issue)
+
+## The ask
+
+Raised in review of PR #521. A handful of recipient/permission queries express
+"is this user an admin-or-event-manager" with a repeated two-part `Q` filter, e.g.
+`backend/notifications/service.py:159`:
+
+```python
+UserModel.objects.filter(
+    Q(roles__name="admin", roles__is_default=True)
+    | Q(roles__permissions__contains=PermissionKey.MANAGE_EVENTS)
+).distinct()
+```
+
+The reviewer asked whether we can stop repeating this `admin role OR permission`
+check. One idea floated was an `is_admin` flag, true only for the role literally
+named `admin`, so call sites do a clean `is_admin` check. This is an
+**investigation, not a directed change** — output is a recommendation, no
+production behavior change in this issue.
+
+Acceptance criteria:
+1. Enumerate every site that hand-rolls the `admin role OR <permission>` filter.
+2. Evaluate options (derived property/manager, stored field, permission-only, leave as-is) with trade-offs.
+3. Recommend one approach with a migration/refactor sketch and blast radius.
+4. No production behavior change here.
+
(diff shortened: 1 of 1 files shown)


leahpeker:
Spike complete. Findings captured; implementation follow-up tracked in #943 (add `User.objects.with_permission()` helper, dedupe the three notification recipient filters). Closing this spec-only PR unmerged — the spec doc has served its purpose as the investigation record.

PR #552, opened 2026-06-25
feat(events): add curated event tags — backend (Issue 511)

leahpeker:
## Summary

Part 1 of 2 for Issue 511. Frontend in #<next>.

- New `EventTag` model (uuid pk, unique `name`, auto-slug) + `Event.tags` M2M (migration `0059`)
- Public `GET /event-tags/` lists the curated set (read-only; id/name/slug)
- `tag_ids` on `EventIn`/`EventPatchIn`; `tags` on `EventOut`/`EventListOut`
- Tag assignment reuses the existing event-edit gate (creator / co-host / `manage_events`); unknown tag ids silently dropped (admin-curated set)
- `tags` prefetched on every event queryset — no N+1 on the list path (regression-guarded by a constant-query test)
- Django admin for `EventTag` + tag picker/filter on `EventAdmin`

## Test plan

- [ ] `make agent-ci` (tests pass, incl. 12 new event-tag tests + N+1 guard)
- [ ] In Django admin, create a couple of `EventTag`s
- [ ] `GET /api/community/event-tags/` works while logged out


backend/tests/test_event_management.py
```diff
@@ -386,7 +386,7 @@ def test_event_patch_fields_match_model(self):
         """All EventPatchIn fields (except M2M and transient fields) must exist on Event model."""
         from community.api import EventPatchIn
 
-        non_model_fields = {"co_host_ids", "status", "notify_attendees"}
+        non_model_fields = {"co_host_ids", "tag_ids", "status", "notify_attendees"}
```
leahpeker, line 384: why are these non-model fields?

Issue #493, opened 2026-06-23
feat(public-rsvp): public RSVP submit endpoint + dedup + confirmation email [stage 3/8]


leahpeker:
Part of epic #490. **Read the full spec first:** [docs/superpowers/specs/2026-05-15-public-rsvp-official-events-design.md](https://github.com/ProteinDeficientsAnonymous/pda/blob/main/docs/superpowers/specs/2026-05-15-public-rsvp-official-events-design.md) — sections **API surface → POST /api/public/events/{event_id}/rsvp/**, **Error handling & edge cases**, **Email templates**.

## Dependencies
- #491 stage 1 (`is_member`, `members()`)
- #492 stage 2 (`NonMemberRsvpToken` + issuance helper)

## Scope
`POST /api/public/events/{event_id}/rsvp/` — **no auth**, rate limit `5/h` per IP (`key_func=lambda r: r.META.get("REMOTE_ADDR", "anon")`, see `.claude/rules/rate-limiting-and-input-validation.md`).

Body: `{ name (max 100), email (EmailStr), phone_number (E.164), status (RSVPStatus), has_plus_one (bool), honeypot? }`. Add `max_length` on the schema. Honeypot follows `_honeypot_decoy_response` in `backend/community/_join_requests.py`.

### Behavior (spec lines 90–103)
1. Validate event is `OFFICIAL` + `status=ACTIVE` + `rsvp_enabled=True` + not past + not cancelled + `visibility=PUBLIC`. Anything else → **404 `Code.Event.NOT_FOUND`** (don't leak non-public events).
2. Dedup by phone AND email (read the spec's collision table carefully — lines 92–98):
   - Either match is `is_member=True` → **409 `Code.RSVP.MEMBER_CONTACT_MUST_SIGN_IN`** (new code; add it).
   - Phone matches non-member, email doesn't → reuse phone row; save email only if blank (never overwrite).
   - Email matches non-member, phone doesn't → reuse email row.
   - Both match same non-member row → reuse, no change.
   - Phone & email match *different* non-member rows → **phone wins**, don't update email, log warning.
   - Neither → `get_or_create` a `User(is_member=False, set_unusable_password())` inside the txn (race-safe).
3. Reuse `_apply_rsvp_in_transaction` (`backend/community/_event_rsvps.py`) verbatim — capacity, plus-ones, auto-waitlist, promotion all reused.
4. Issue a fresh `NonMemberRsvpToken`.
[...]

PR #557, opened 2026-06-25
feat(public-rsvp): enforce non-members never hold a role (Issue 525)

### LINKED ISSUE #525: feat(public-rsvp): enforce non-members can never hold a role + trustworthy role counts
Part of epic #490. Follow-up from PR #521 (stage 1 — `is_member` / `members()`).

## Problem
Non-members (public-RSVP `User` rows, `is_member=False`) must **never** hold a role. PR #521 added `User.objects.members()` and converted member-facing surfaces, but the invariant "a non-member cannot be assigned a role" is only partially enforced, and the role **member-count surfaces do not defend against it**.

### Current state (as of PR #521)
- ✅ `update_user_roles` (`backend/users/_management.py`) already fetches the target via `User.objects.members().get(...)`, so it 404s when assigning roles to a non-member. Good.
- ❌ **`list_roles`** (`backend/users/_roles.py:36`) computes `user_count=Count("users", filter=Q(users__archived_at__isnull=True))` — **no `is_member` filter**. If a non-member ever holds a role (via a bug or a future code path), the per-role count shown in the role-management UI is silently inflated.
- ❌ **`delete_role`** (`backend/users/_roles.py:176`) computes `affected_user_count = role.users.filter(archived_at__isnull=True).count()` — same gap, logged to the audit trail.

## Decision
The clean fix is to **enforce the invariant at write time** so the read-side counts can stay simple and trustworthy:

1. **Frontend (members list / role editor):** ensure non-members cannot be offered role assignment in the UI (they shouldn't appear in the members list at all — verify after stage 3 creates non-members).
2. **Backend guard:** confirm every role-assignment path rejects non-members (audit `user.roles.add(...)` / `user.roles.set(...)` call sites — `_helpers.py`, `_management.py`, `seed.py`). `update_user_roles` is covered; verify `_create_user_with_role` and `bulk_create_users` can't attach a role to an `is_member=False` user.
3. **Defense-in-depth on counts (optional):** add `users__is_member=True` to the `list_roles` and `delete_role` counts so a bug elsewhere can't inflate them. Belt-and-suspenders given (1)+(2).

## Acceptance criteria
- No code path can assign a role to an `is_member=False` user (covered by a test).
- Role member counts in `list_roles` / `delete_role` reflect members only (either by the write-time invariant alone, or with the explicit `is_member` filter as backup).
- Frontend members list cannot offer role assignment to a non-member.

## Notes
- This only matters once a non-member-creating path lands (stage 3, #493). Until then the counts are trivially correct (no non-members exist). Assert the queryset filters directly rather than relying on real non-member data.

## Verify
`make agent-ci` before done.

leahpeker:
## Summary
Closes #525. Part of epic #490, follow-up to PR #521 (stage 1: `is_member` / `members()`).

- **Write-time invariant:** new `reject_role_for_non_member` `m2m_changed` (`pre_add`) receiver on `User.roles.through` raises `ValueError` if any non-member (`is_member=False`) would be assigned a role. Call-site-independent — fires on `user.roles.add/set/create` and the reverse `role.users.add/set` from every path (API, seed, the superuser `post_save`, and any future code), in both `reverse=True`/`False` directions.
- **Trustworthy counts (defense-in-depth):** `list_roles` and `delete_role` member-counts now filter on `users__is_member=True`, so a bug elsewhere can never silently inflate per-role counts even if a non-member ever slipped a role at the DB level.
- Existing `update_user_roles` already 404s non-members via `User.objects.members()`; that gate is the expected first line — reaching the guard means a caller skipped it, so the `ValueError` surfaces as a loud 500 by design.

The frontend members-list acceptance point is satisfied trivially today: no non-member-creating path exists yet (lands in stage 3, #493), so non-members can't appear in the list to be offered roles.

## Test plan
- [x] `TestNonMemberCannotHoldRole`: forward `add`/`set`, reverse `users.add`/`users.set` all reject non-members; member assignment + member role-swap via `set()` still work; `update_user_roles` 404s a non-member; `list_roles` count excludes a DB-forced non-member.
- [x] `make agent-ci` green (1026 backend tests, typecheck, complexity, schema checks; frontend lint/format/typecheck/types-check).

### DIFF
--- a/backend/tests/test_role_management.py
+++ b/backend/tests/test_role_management.py
@@ -164,3 +164,88 @@ class TestRoleManagementAPI:
         assert response.status_code == 200
         popular = next(r for r in response.json() if r["name"] == "popular")
         assert popular["user_count"] == 1
+
+
+@pytest.mark.django_db
+class TestNonMemberCannotHoldRole:
+    """Issue 525: a non-member (is_member=False) must never hold a role, and the
+    role member-counts must reflect members only."""
+
+    def _make_non_member(self):
+        from users.models import User
+
+        # is_member=False overrides the conftest autouse default (is_member=True).
+        return User.objects.create_user(
+            phone_number="+12025559001",
+            display_name="Non Member",
+            is_member=False,
+        )
+
+    # The guard raises inside the m2m operation's atomic block, which marks the
+    # surrounding transaction as broken — so we cannot query the DB after the
+    # raise in the same test. pytest.raises is sufficient proof the write aborted.
+    def test_roles_add_rejects_non_member(self):
+        non_member = self._make_non_member()
+        role = Role.objects.create(name="contributor", permissions=[])
+        with pytest.raises(ValueError, match="non-member"):
+            non_member.roles.add(role)
+
+    def test_roles_set_rejects_non_member(self):
+        non_member = self._make_non_member()
+        role = Role.objects.create(name="contributor", permissions=[])
+        with pytest.raises(ValueError, match="non-member"):
+            non_member.roles.set([role])
+
+    def test_reverse_users_add_rejects_non_member(self):
+        non_member = self._make_non_member()
+        role = Role.objects.create(name="contributor", permissions=[])
+        with pytest.raises(ValueError, match="non-member"):
+            role.users.add(non_member)
+
+    def test_reverse_users_set_rejects_non_member(self):
+        non_member = self._make_non_member()
+        role = Role.objects.create(name="contributor", permissions=[])
+        with pytest.raises(ValueError, match="non-member"):
+            role.users.set([non_member])
+
+    def test_member_can_still_hold_role(self, test_user):
+        # The guard must not block legitimate member assignment.
+        role = Role.objects.create(name="contributor", permissions=[])
+        test_user.roles.add(role)
+        assert test_user.roles.filter(name="contributor").exists()
+
+    def test_member_can_swap_roles_via_set(self, test_user):
+        # Regression guard: set() emits pre_remove alongside pre_add, so a member
+        # replacing one role set with another must not trip the membership check.
+        first = Role.objects.create(name="first", permissions=[])
+        second = Role.objects.create(name="second", permissions=[])
+        test_user.roles.set([first])
+        test_user.roles.set([second])
+        names = set(test_user.roles.values_list("name", flat=True))
+        assert names == {"second"}
+
+    def test_update_user_roles_404s_for_non_member(self, api_client, manage_users_headers):
+        non_member = self._make_non_member()
+        member_role = Role.objects.get(name="member")
+        response = api_client.patch(
+            f"/api/auth/users/{non_member.id}/roles/",
+            {"role_ids": [str(member_role.id)]},
+            content_type="application/json",
+            **manage_users_headers,
+        )
+        assert response.status_code == 404
+        assert_error_code(response, Code.User.NOT_FOUND)
+
+    def test_list_roles_count_excludes_non_member(
+        self, api_client, manage_users_headers, test_user
+    ):
+        # Force a non-member onto a role at the DB level (bypassing the guard)
+        # to prove the read-side count defends itself anyway.
+        role = Role.objects.create(name="counted", permissions=[])
+        test_user.roles.add(role)
+        non_member = self._make_non_member()
+        role.users.through.objects.create(user=non_member, role=role)
+        response = api_client.get("/api/auth/roles/", **manage_users_headers)
+        assert response.status_code == 200
+        counted = next(r for r in response.json() if r["name"] == "counted")
+        assert counted["user_count"] == 1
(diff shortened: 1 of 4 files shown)


leahpeker:
⚠️ Merge conflict with `main` needs manual resolution.

A rebase onto `main` hits a content conflict in `backend/users/models.py`. Both sides **added independent blocks at the same location** (right before `assign_admin_role_to_superuser`):

- **main** added the `NonMemberRsvpToken` model + `NON_MEMBER_RSVP_TOKEN_TTL_DAYS` constant.
- **this PR** added the `reject_role_for_non_member` `m2m_changed` receiver.

The two additions don't overlap logically — the resolution is to **keep both blocks**. Auto-resolution was skipped because `models.py` is a logic file (outside the safe lock-file / generated-file / import-union categories), so the rebase was aborted and the branch left untouched. A human should rebase and union both blocks.

backend/users/models.py
```diff
@@ -163,6 +163,35 @@ def create_for_user(
         )
 
 
+@receiver(m2m_changed, sender=User.roles.through)
+def reject_role_for_non_member(sender, instance, action, reverse, pk_set, **kwargs):
+    """Enforce the invariant: a non-member (is_member=False) can never hold a role.
+
+    This is the call-site-independent guard for Issue 525 — it fires on
```
leahpeker: this whole block is wayyyy too verbose

leahpeker: tightened the docstring (and the inline comments) considerably in f0cc859 — kept only the essential 'why' (fix the caller, don't catch) and the reverse/forward orientation note.

leahpeker:
resolved: rebased onto latest `main` and unioned both blocks in `backend/users/models.py` (`main`'s `NonMemberRsvpToken` + `NON_MEMBER_RSVP_TOKEN_TTL_DAYS`, this PR's `reject_role_for_non_member` receiver) — they don't overlap, so both are kept verbatim. `make agent-ci` passes; pushed via `--force-with-lease`.

backend/users/models.py
```diff
@@ -243,6 +243,26 @@ def resolve_user(cls, token: str) -> "User | None":
...
+    """Enforce the Issue 525 invariant: a non-member can never hold a role.
+
+    Fires on every role-assignment path (``user.roles.add``, ``role.users.add``,
+    etc.). Callers should gate on membership first; reaching this guard means one
+    didn't, so the ``ValueError`` surfaces as a loud 500 — fix the caller, don't
+    catch it.
+    """
+    if action != "pre_add" or not pk_set:
+        return
+    # reverse=True: instance is the Role, pk_set holds User PKs.
```
leahpeker: get rid of these comments and name reverse better / more meaningful

backend/tests/test_role_management.py
```diff
@@ -164,3 +164,88 @@ def test_list_roles_includes_user_count(self, api_client, manage_users_headers,
...
+        from users.models import User
+
+        # is_member=False overrides the conftest autouse default (is_member=True).
+        return User.objects.create_user(
+            phone_number="+12025559001",
+            display_name="Non Member",
+            is_member=False,
+        )
+
+    # The guard raises inside the m2m operation's atomic block, which marks the
```
leahpeker: limit comment to 1 line max

backend/users/_roles.py
```diff
@@ -173,7 +179,9 @@ def delete_role(request, role_id: str):
     if role.name in PROTECTED_ROLE_NAMES:
         raise_validation(Code.Role.PROTECTED_CANNOT_DELETE, status_code=400, role_name=role.name)
     role_name = role.name
-    affected_user_count = role.users.filter(archived_at__isnull=True).count()
+    # Members only — see the matching note in list_roles. Keeps the audited
```
leahpeker: limit comment to 1 line max

backend/users/models.py
leahpeker: done in 7cc1220 — dropped the comments and replaced the reverse branch with an `instance_is_role` local. (the `reverse` param name itself is fixed by django's m2m_changed signal signature, so it can't be renamed.)

backend/tests/test_role_management.py
leahpeker: done in 7cc1220 — collapsed to one line.

backend/users/_roles.py
leahpeker: done in 7cc1220 — collapsed to one line.

leahpeker:
the merge-conflict in `backend/users/models.py` was already resolved earlier (both blocks unioned, see the follow-up comment above). no conflict remains on the current branch — `git merge-base` is clean and `make agent-ci` passes on the latest commit (7cc1220).

leahpeker:
resolved: rebased onto latest `main`. the only remaining conflict was in `.claude/task.md` (a tracking file) where `main` had advanced with an unrelated task — kept this PR's #525 tracking content. `backend/users/models.py` auto-merged cleanly (both the `NonMemberRsvpToken`/`NON_MEMBER_RSVP_TOKEN_TTL_DAYS` block from main and this PR's `reject_role_for_non_member` receiver are present). `make agent-ci` passes (1026 backend + 547 frontend tests, typecheck, complexity, schema, lint, format). pushed via `--force-with-lease` (247be37). no merge conflict remains; PR is mergeable pending review.

PR #561, opened 2026-06-25
fix(profile): stamp photoUpdatedAt after upload so avatar refreshes (Issue 551)

### LINKED ISSUE #551: fix(profile): profile photo doesn't update in UI after upload without reload
## Context

After uploading a new profile photo, the avatar in the UI doesn't update until the page is reloaded. The backend does not return a `photo_updated_at` field, so `photoUpdatedAt` is always `null` in the frontend user object. The cache-buster URL in `AvatarUpload.tsx` (`?v=${encodeURIComponent(user.photoUpdatedAt)}`) never changes, so the browser serves the stale cached image.

## What needs to happen

After a successful upload, stamp `photoUpdatedAt` client-side with the current ISO timestamp so the `img` src changes and the browser fetches the newly uploaded photo without a reload.

- `frontend/src/auth/store.ts` — in `uploadProfilePhoto` (~line 119), set `user: { ...user, photoUpdatedAt: new Date().toISOString() }` before calling `set()`
- `frontend/src/screens/settings/AvatarUpload.tsx` — no changes needed; the cache-buster at lines 31–35 already uses `photoUpdatedAt`

## Acceptance criteria

- [ ] Uploading a new profile photo immediately updates the avatar in the UI without a page reload
- [ ] The cache-busting `?v=` param on the avatar `img` src changes after each upload

leahpeker:
## Summary

Closes #551

- After uploading a profile photo the avatar didn't update until a page reload. The backend `UserSchema` omits `photo_updated_at`, so `photoUpdatedAt` is always `null` in the frontend user object and the cache-buster `?v=${photoUpdatedAt}` on the avatar `img` src never changes — the browser serves the stale cached image.
- Fix: in the auth store's `uploadProfilePhoto` action, stamp `photoUpdatedAt` client-side with the current ISO timestamp (`new Date().toISOString()`) after the upload, so the img src changes and the new photo loads without a reload.
- `AvatarUpload.tsx` (and `BottomNav.tsx`) already consume `photoUpdatedAt` for the cache-buster — no changes needed there.

## Test plan

- [x] `make agent-frontend-test` — full Vitest suite passes
- [x] `make agent-frontend-lint` — eslint + prettier clean
- [x] `make agent-frontend-typecheck` — tsc clean
- [x] New unit test asserts `uploadProfilePhoto` stamps a fresh `photoUpdatedAt` even when the API returns `photoUpdatedAt: null` (mirrors real backend behavior)
- [ ] Manual: upload a new profile photo and confirm the avatar updates immediately without reloading; verify the `?v=` param on the img src changes


frontend/src/auth/store.ts
```diff
@@ -118,7 +118,9 @@ export const useAuthStore = create<AuthState>((set, get) => ({
 
   async uploadProfilePhoto(file) {
     const user = await authApi.uploadProfilePhoto(file);
-    set({ user });
+    // Backend omits photo_updated_at, so without this stamp the cache-buster
```
leahpeker: would it make more sense to not have the backend omit the photo_updated_at?

leahpeker: good call — done. instead of stamping `photoUpdatedAt` client-side, the backend no longer omits it:

- added a `User.photo_updated_at` field (+ migration `0030`)
- stamp it on photo upload, clear it on delete
- expose `photo_updated_at` in `UserOut`

the frontend now receives a real server timestamp on the upload response, so the cache-buster `?v=` changes naturally and the avatar refreshes without a reload. removed the `new Date().toISOString()` stamp in `store.ts` — the backend value is the source of truth now. types regenerated; backend + frontend ci green.

backend/users/models.py
```diff
@@ -84,6 +84,10 @@ class User(AbstractUser):
     calendar_token = models.CharField(max_length=64, blank=True, default="", db_index=True)
     bio = models.CharField(max_length=500, blank=True, default="")
     profile_photo = models.ImageField(upload_to="profile_photos/", blank=True)
+    # Stamped whenever profile_photo is uploaded; surfaced on /me/ as
```
leahpeker: get rid of this comment

backend/tests/test_photos.py
```diff
@@ -77,8 +77,12 @@ def test_upload_photo(self, api_client, member):
         assert response.status_code == 200
         data = response.json()
         assert data["profile_photo_url"] != ""
+        # The response carries a fresh server timestamp the frontend uses to
```
leahpeker: get rid of the ocmment

PR #587, opened 2026-06-28
feat(public-rsvp): extend existing non-member rsvp token on return (Issue 555)

leahpeker:
## Summary
- Add `NonMemberRsvpToken.issue_or_extend()`: a returning non-member with a valid token gets its `expires_at` bumped to now + 90 days while keeping the **same token string**, so a `/my-rsvps` link saved from a previous email keeps resolving. A fresh token is only minted when no valid one exists (expired or revoked). Members are rejected, matching `issue()`.
- Update the model docstring to reflect extend-vs-create (tokens are no longer always issued fresh on every RSVP).
- The public RSVP submission endpoints (stages 3/4) that will call this — and the `User` get-or-create-by-phone for recognizing returning non-members (AC #1) — are not built yet. This lands the model-layer extend-vs-create logic (AC #2/#3) and its tests ahead of them. No schema change, so no migration.

Part of the public-RSVP epic (#490).

## Test plan
- [ ] `cd backend && uv run pytest tests/test_nonmember_rsvp_token.py` — `TestIssueOrExtend` covers: first call issues fresh, valid token extended in place (same pk + token, old URL still resolves), expired → fresh, revoked → fresh (revoked stays revoked), member rejected.
- [ ] `make agent-ci` — backend + frontend green.

### DIFF
--- a/backend/users/models.py
+++ b/backend/users/models.py
@@ -175,9 +175,11 @@ class NonMemberRsvpToken(models.Model):
 
     This token NEVER logs the user in — it is entirely separate from
     MagicLoginToken (different table, different validation path, narrower scope).
-    It grants scoped read/write to one non-member's RSVPs only. Issued (fresh)
-    on every successful non-member RSVP and delivered by email; old tokens stay
-    valid until they expire. Revoked when the user converts to a member.
+    It grants scoped read/write to one non-member's RSVPs only. On a non-member's
+    RSVP we EXTEND their existing valid token (see issue_or_extend) rather than
+    minting a new one, so a link saved from a previous email keeps working; a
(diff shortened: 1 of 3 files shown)


backend/users/models.py
```diff
@@ -228,6 +230,27 @@ def issue(cls, user: "User") -> "NonMemberRsvpToken":
             expires_at=timezone.now() + timedelta(days=NON_MEMBER_RSVP_TOKEN_TTL_DAYS),
         )
 
+    @classmethod
+    def issue_or_extend(cls, user: "User") -> "NonMemberRsvpToken":
+        """Return a usable non-member RSVP token, reusing one when possible.
```
leahpeker, line 235: this is way too verbose

PR #589, opened 2026-06-28
fix(photos): drop unsupported heic uploads and surface crop decode errors (Issue 505)

### LINKED ISSUE #505: photo upload didnt work

leahpeker:
## Summary

Closes #505

A user reported event photo upload "didn't work" (macOS Chrome). Root cause: the app advertised heic/heif support, but the client-side crop pipeline decodes images via a browser ``/canvas, which **cannot decode heic/heif** (the Apple Photos default). A user could pick a `.heic`, pass validation, then land in a crop dialog whose `` never loaded — leaving the "save" button permanently disabled with **no error**. A silent "photo upload didn't work."

- Remove `image/heic` / `image/heif` from the frontend accept/validation lists (`EventFormPhoto.tsx`, `AvatarUpload.tsx`) and the backend allowed-type sets (`_event_schemas.py`, `users/_auth.py`) so the contract matches what the crop pipeline can actually process.
- Add an `onError` handler + `role="alert"` message to `ImageCropDialog` so any undecodable image surfaces a clear error instead of a dead button.
- Tests: backend rejects heic on both photo endpoints; frontend surfaces the decode error and keeps save disabled.

## Test plan

- [ ] `make agent-ci` passes (verified locally: 1005 backend + 547 frontend tests green)
- [ ] upload a jpeg/png/webp/gif event photo → crops and persists
- [ ] pick a `.heic` → blocked at file selection with a clear "pick a jpeg, png, webp, or gif image" message
- [ ] if an undecodable image still reaches the crop dialog → "couldn't read that image…" alert shows and save stays disabled

### DIFF
--- a/backend/community/_event_schemas.py
+++ b/backend/community/_event_schemas.py
@@ -404,11 +404,12 @@ class EventPatchIn(BaseModel):
 
 
 _MAX_EVENT_PHOTO_SIZE = 10 * 1024 * 1024  # 10 MB
+# heic/heif are intentionally excluded: the frontend crops via a browser canvas
+# that can't decode them, so such uploads can only arrive via a hand-crafted
+# request and would never round-trip through the real UI (issue 505).
 _ALLOWED_IMAGE_TYPES = {
     "image/jpeg",
     "image/png",
     "image/webp",
     "image/gif",
-    "image/heic",
-    "image/heif",
 }
--- a/backend/tests/test_photos.py
+++ b/backend/tests/test_photos.py
@@ -102,6 +102,12 @@ class TestProfilePhoto:
         response = api_client.post("/api/auth/me/photo/", {"photo": big}, **_auth(member))
         assert response.status_code == 400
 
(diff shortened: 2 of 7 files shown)


leahpeker:
i dont want to drop heic uploads - this is not the correct solution

PR #603, opened 2026-07-08
fix(profile): cache-bust avatar after upload via server photo_updated_at (Issue 551)

### LINKED ISSUE #551: fix(profile): profile photo doesn't update in UI after upload without reload

leahpeker:
## Summary
- Fixes the profile avatar not refreshing in the UI after upload until a page reload.
- Adds a real server-side `photo_updated_at` `DateTimeField` on `User` (migration `0030`) instead of the issue's suggested client-side timestamp stamp — the server value survives reloads, stays consistent across every client/device, and is correctly nulled on delete.
- `upload_photo` stamps `timezone.now()` (single `save(update_fields=[...])`); `delete_photo` nulls it. `UserOut` returns it as an ISO string; the frontend maps `photo_updated_at → photoUpdatedAt`, and the existing cache-buster `?v=${photoUpdatedAt}` in `AvatarUpload.tsx` and `BottomNav.tsx` picks up the new value automatically.

Closes #551

## Test plan
- [ ] Upload a new profile photo and confirm the avatar updates immediately (no reload) in both the settings screen and bottom nav.
- [ ] Delete the profile photo and confirm the avatar clears.
- [ ] `backend/tests/test_photos.py::TestProfilePhoto` — asserts `photo_updated_at` is set on upload and nulled on delete (7 passed).
- [ ] `frontend/src/auth/store.test.ts` — asserts the store persists the server-stamped `photoUpdatedAt`.

### DIFF
--- a/frontend/src/auth/store.test.ts
+++ b/frontend/src/auth/store.test.ts
@@ -177,9 +177,6 @@ describe('useAuthStore', () => {
 
   describe('uploadProfilePhoto', () => {
     it('stores the server-stamped photoUpdatedAt so the cache-buster changes', async () => {
-      // The backend now returns a fresh photo_updated_at on upload, which the
-      // store persists verbatim. The cache-buster ?v= param keys off it, so the
-      // avatar refreshes without a reload — no client-side stamping needed.
       useAuthStore.setState({ status: 'authed', user: mockUser, accessToken: 'tok-abc' });
       const serverStamp = '2026-06-25T12:00:00+00:00';
       vi.mocked(authApi.uploadProfilePhoto).mockResolvedValueOnce({
--- a/frontend/src/auth/store.ts
+++ b/frontend/src/auth/store.ts
@@ -120,9 +120,7 @@ export const useAuthStore = create<AuthState>((set, get) => ({
     // The upload response carries a fresh server-stamped photoUpdatedAt, which
     // drives the cache-buster ?v= param so the avatar refreshes without a reload.
     const user = await authApi.uploadProfilePhoto(file);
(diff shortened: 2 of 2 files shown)


frontend/src/api/types.gen.ts
leahpeker, line 3434: remove comments on these

frontend/src/auth/store.ts
leahpeker: remove comment

frontend/src/auth/store.test.ts
leahpeker: remove comment

PR #650, opened 2026-07-10
fix(profile): bound image crop ops so upload cannot wedge on saving (Issue 580)

### LINKED ISSUE #580: Profile pic save got stuck mobile
```
Got stuck showing saving… text but even I reloaded the new pic was there
```

## Metadata

- **Route:** `/profile`
- **User Agent:** `Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.5 Mobile/15E148 Safari/604.1 Brave`

leahpeker:
## Overview

Fixes #580 — on mobile (iPhone Safari/Brave), saving a profile photo could get stuck on "saving…" forever, even though the photo saved server-side.

## Root cause

The crop dialog's `handleSave` does `setSaving(true)` then `await cropImage(...)` inside a `try/finally`. `cropImage` awaits two callback-based promises — `loadImage` (Image.onload) and `canvasToBlob` (canvas.toBlob) — **neither of which had a timeout**. On iOS Safari, `Image.onload` / `canvas.toBlob` can fail to fire their callback for certain photo-library images, so the promise never settles, `handleSave`'s `finally` never runs, and the UI stays on "saving…" with no recovery.

(The photo appearing after reload is consistent with a prior/duplicate attempt having reached the server while the client was wedged.)

## Fix

Wrap both `loadImage` and `canvasToBlob` in a 15s timeout via a shared `withTimeout` helper in `cropImage.ts`. Now the promise always settles: on a wedge it rejects, `handleSave`'s `finally` clears the saving state, and the user sees an actionable error instead of an infinite spinner.

## Tests

- New `cropImage.test.ts`: asserts `cropImage` **rejects** (not hangs) when `canvas.toBlob` never invokes its callback, plus a happy-path case. Verified the timeout test **hangs/fails without the guard**.
- Full frontend suite green: `629 passed / 1 skipped`, typecheck + eslint + prettier clean.

## Note

This is a defensive fix at the crop layer that covers the whole class of "callback never fires" wedges (the same `cropImage` is used by event-photo cropping too). It does not change the happy path.

### DIFF
--- a/frontend/src/components/ImageCropDialog.test.tsx
+++ b/frontend/src/components/ImageCropDialog.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, render, screen } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import type { ReactNode } from 'react';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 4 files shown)


leahpeker:
self-review follow-up (pushed aad012c): the timeout guard stopped the infinite 'saving…' hang, but `handleSave` had no catch and runs as `void handleSave()`, so a crop failure reverted the button silently with no feedback. added a catch that surfaces a friendly error in the dialog. also moved the src object-url revoke into a useEffect unmount cleanup (the error path previously leaked it), dropped the now-empty handleCancel wrapper, and trimmed the multi-line comment to one line. added an error-path test. full ci green (630 passed).

PR #654, opened 2026-07-10
fix(events): wrap long text so event detail can't overflow on mobile (Issue 611)

### LINKED ISSUE #611: Ui
```
Event detail page takes up full width
```

## Metadata

- **Route:** `/events/723fc561-7950-4eb4-8bd9-375b0ed80c2a`
- **User Agent:** `Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.5 Mobile/15E148 Safari/604.1 Brave`

leahpeker:
## Overview

Fixes #611 — on mobile the event detail page rendered with the content boxes flush to both screen edges; zooming out snapped it back to the normal constrained layout. That "zoom fixes it" behaviour is the tell-tale sign of **horizontal overflow**, not a missing max-width (the page already uses the standard `ContentContainer` = `max-w-3xl`, same as every other screen).

## Root cause

The event description is rendered with `whitespace-pre-wrap` but **no word-breaking**, and `linkifyText` turns urls into `<a>` elements. A long unbroken url (partiful / whatsapp / maps links are common in descriptions) is a single token with no break opportunity, so it stretched the content column past the phone's viewport width. The browser then scaled the whole page down to fit, leaving the boxes touching both edges until the user manually zoomed out.

## Fix

Add `break-words` (`overflow-wrap: break-word`) to the description paragraph — and to the title `<h1>`, which had the same exposure for a long unbroken title — so long tokens wrap instead of forcing overflow.

## Tests

- New regression test in `EventDetailScreen.test.tsx`: renders a description containing a 200-char url and asserts the description paragraph carries `break-words`. Verified it **fails without the fix**.
- Full frontend suite green: `628 passed / 1 skipped`, typecheck + eslint + prettier clean.

## Note

jsdom has no layout engine, so the test guards the wrapping *class* rather than measuring pixels. The fix is a CSS-only change with no logic impact.

### DIFF
--- a/frontend/src/screens/events/EventDetailScreen.test.tsx
+++ b/frontend/src/screens/events/EventDetailScreen.test.tsx
@@ -173,10 +173,13 @@ describe('EventDetailScreen', () => {
     } as ReturnType<typeof useEvent>);
     renderScreen();
 
-    const link = screen.getByRole('link', { name: new RegExp(longUrl.slice(8, 40)) });
+    const namePattern = new RegExp(longUrl.slice(8, 40).replace(/\./g, '\\.'));
+    const link = screen.getByRole('link', { name: namePattern });
     const paragraph = link.closest('p');
     expect(paragraph).not.toBeNull();
-    expect(paragraph).toHaveClass('break-words');
+    // break-words alone doesn't reliably break a spaceless token in WebKit;
+    // overflow-wrap:anywhere is what actually forces the break (see RichEditor).
+    expect(paragraph).toHaveClass('break-words', '[overflow-wrap:anywhere]');
   });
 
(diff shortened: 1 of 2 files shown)


leahpeker:
self-review follow-up (pushed c3d0fc1): `break-words` alone doesn't reliably break a spaceless long token in mobile safari — the exact target of this issue. switched to `break-words [overflow-wrap:anywhere]` to match the existing codebase convention (RichEditor.tsx uses the same pairing; admin dialogs use break-all for long urls/codes). applied to the title too, and the test now asserts both classes. full ci green.

Issue #604, opened 2026-07-08
clarify event type vs. visibility taxonomy (they're conflated)


leahpeker:
## Context

Raised in review of [#569](https://github.com/ProteinDeficientsAnonymous/pda/pull/569) ([thread](https://github.com/ProteinDeficientsAnonymous/pda/pull/569#discussion_r3544362294)). The public-RSVP eligibility check (`_load_public_rsvp_event`) requires an event to be **both** `event_type == OFFICIAL` **and** `visibility == PUBLIC`. That's a symptom of a deeper conflation: official events must always be public, so `event_type` and `visibility` overlap and are confusing.

## The problem

Today the two axes are:

- **`event_type`**: `official` / (club) / members-only / etc.
- **`visibility`**: `public` / `invite_only` / `members`

But their meanings overlap and are hard to explain to users:

- **public** — shows up publicly, but you still need to be a member to RSVP
- **invite only** — not on your calendar unless you've been invited
- **members** — shows up semi-publicly, but only to members
- **official** — public *and* designated official

The confusing part: all types *except* "official" are effectively members-only to RSVP. So the type and visibility dimensions are tangled.

## Proposed direction (to discuss)

Split the two concerns cleanly:

- **type**: `official` / `club` / `members only`
- **visibility**: `invite only` vs. not invite only

This removes the redundant "official must be public" coupling and gives users two orthogonal, understandable choices.

## Scope

Design + clarification issue — needs decisions on the final type/visibility taxonomy, migration of existing events, and UI copy before implementation. Out of scope for #569.

leahpeker:
🔧 Draft PR created: https://github.com/ProteinDeficientsAnonymous/pda/pull/659

Explored the event type vs. visibility conflation and committed a findings spec (`docs/superpowers/specs/2026-07-10-event-type-visibility-taxonomy-explore.md`). No code changed — this is a design/decision doc.

**Three things the issue framing didn't capture:**
1. There's **no `club`/`members-only` *type*** today — `event_type` is only `official`/`community`. The proposed type split would be *new* taxonomy, not a re-slice.
2. `PageVisibility` is a **shared** enum (Event *and* EditablePage) — a clean redesign should split out an event-specific enum, not mutate the shared one.
3. The **frontend already merges both axes into one "who can see it" dropdown** (`visibilityChoiceToFields`), so the conflation is now mostly a backend-model + copy problem. The user pain shows up as near-identical `public` vs `official` helper text.

[...]

PR #692, opened 2026-07-11
feat: 604 - add club event type + tag_club_event permission

leahpeker:
## Overview

Follow-up to #687 (now merged). Adds a **club** event type and a dedicated `tag_club_event` permission so event planners with the right role can tag an event as a PDA club event — separate from official, and gated independently.

- **Backend** — new `EventType.CLUB` (public-only, like official) + `PermissionKey.TAG_CLUB_EVENT`. The per-type permission check is unified into `_enforce_type_tag_permission` (maps event type → required permission), replacing the copy-pasted official gates in create + update. Club joins `_PUBLIC_ONLY_TYPES`, inheriting the public-only invariant. Migration `0064` alters the `event_type` choices (no data change).
- **Frontend** — "make it a pda club event" toggle in `EventFormType`, shown to users with `tag_club_event`. Because `event_type` is single-valued, the official and club toggles are mutually exclusive; turning either on forces + locks public visibility. Adds the `Permission.TagClubEvent` mirror + its `RoleFormDialog` label, `EventType.Club`, a rose/magenta color (CVD-safe, verified distinct from the other four categories in `eventColors.test.ts`), a club badge on the detail screen, a club chip on the admin list, and a "pda club" calendar filter option.

Completes the type/visibility separation for #604.

## Test plan

- [x] Backend `test_event_club.py` (new): club requires `tag_club_event` (403 without), club + non-public rejected (400), club visible to anon in list
- [x] `test_event_visibility.py` split to stay under the 500-line limit; official tests unchanged and green
- [x] Frontend: `EventFormType.test.tsx` covers both toggles + mutual exclusion; `eventColors.test.ts` asserts 5 distinct categories; `event.test.ts` covers the club class
- [x] Live API check on seeded SQLite: admin club+public → 201, club+members_only → 400, member-without-permission → 403
- [ ] Manual UI pass (grant `tag_club_event`, confirm toggle + badge + calendar filter)

## Notes

Depended on #687 (merged); rebased onto main. `event_type` is a plain string in the OpenAPI schema, so adding the `club` value needs no `types.gen.ts` regeneration.


backend/tests/test_event_club.py
```diff
@@ -0,0 +1,83 @@
+"""Tests for club event type — public-only, gated by tag_club_event."""
```
leahpeker: remvoe top of file comments

frontend/src/screens/events/form/EventFormType.tsx
```diff
@@ -6,22 +6,39 @@ interface Props {
   values: EventFormValues;
   onChange: (patch: Partial<EventFormValues>) => void;
   canTagOfficial: boolean;
+  canTagClub: boolean;
 }
 
-export function EventFormType({ values, onChange, canTagOfficial }: Props) {
-  if (!canTagOfficial) return null;
+// event_type is a single value, so the two toggles are mutually exclusive:
```
leahpeker: reduce to 1 line

leahpeker:
addressed the two comment nits in 2267c94 — dropped the test_event_club.py docstring and reduced the EventFormType comment to one line.

PR #707, opened 2026-07-11
fix(rsvp): broadcast event update so capacity UI refreshes live (Issue 637)

### LINKED ISSUE #637: Capacity UI stale under concurrent RSVPs
## Summary
`atCapacity` / "join the waitlist" derived from cached `attendingCount`. Server is correct; UI can mislead until refetch.

## Location
`frontend/src/models/event.ts`, `frontend/src/screens/events/RsvpSection.tsx`

## Found by
Cursor

leahpeker:
## Overview

Makes the backend push a live update when an RSVP changes an event's capacity, so other viewers' UI self-corrects in real time (Issue 637).

`atCapacity` and the "join the waitlist" label are derived from the cached event's `attendingCount`. When **another** user fills the last spot, this client's cache isn't invalidated by that action, so the capacity UI stayed stale (and could mislead) until an unrelated refetch. (The actor's own UI was already correct — the RSVP mutation writes the fresh event back to cache.)

The real-time plumbing already existed and just wasn't wired to RSVPs:
- `broadcast_event_update()` pings the `event_updates` `pg_notify` channel for all event stakeholders (creator + co-hosts + invited + attending/maybe RSVPs).
- The frontend's SSE handler already invalidates `['events']` on the `event_updated` event.
- But only cohost changes called it — the RSVP paths never did.

## Changes

- `backend/community/_event_rsvps.py` (`upsert_rsvp`) and `backend/community/_public_rsvp.py` (`submit_public_rsvp`) — call `broadcast_event_update()` after the RSVP commits, excluding the actor (who already has the fresh event in their response). Both paths already re-load the event with the exact prefetches the broadcast needs.
- Tests: assert each RSVP path calls `broadcast_event_update` (member path also asserts the actor is excluded). Mirrors the existing `_notify_users` patch pattern.

## Notes / scope

- This fixes the **live-refresh** gap. The outcome was always correct regardless — the server auto-waitlists an over-capacity RSVP and returns the corrected event; this makes *other clients'* displayed capacity catch up without a manual refetch.
- `broadcast_event_update` no-ops off Postgres (`connection.vendor` guard), so the SQLite test suite asserts the call, not the `pg_notify` side effect. Consistent with the existing SSE code.

## Verification

- Both new tests fail without the broadcast calls, pass with them — verified by temporarily removing the calls.
- `test_rsvp.py`, `test_event_capacity.py`, `test_public_rsvp.py`, `test_public_rsvp_capacity.py` all pass (88 total).
- ruff lint + ty typecheck clean.

Closes #637

### DIFF
--- a/backend/community/_event_rsvps.py
+++ b/backend/community/_event_rsvps.py
@@ -11,7 +11,6 @@ from django.db import transaction
 from django.utils import timezone
 from ninja import Router
 from ninja.responses import Status
-from notifications.service import broadcast_event_update
 
 from community._event_helpers import (
     _attended_count,
@@ -25,6 +24,7 @@ from community._event_helpers import (
     _no_show_count,
     _not_marked_count,
     _waitlisted_count,
+    broadcast_capacity_change,
     promote_from_waitlist,
 )
 from community._event_schemas import (
@@ -178,14 +178,11 @@ def upsert_rsvp(request, event_id: UUID, payload: RSVPIn):
         target_id=str(event_id),
         details={"status": final_status},
     )
-    event = (
-        Event.objects.select_related("created_by")
-        .prefetch_related("co_hosts", "invited_users", "rsvps__user")
-        .get(id=event_id)
-    )
-    # Live-refresh other viewers' capacity UI (attendingCount / atCapacity).
-    # The actor already has the fresh event from this response, so exclude them.
-    broadcast_event_update(event, exclude_user_ids={str(request.auth.pk)})
+    event = _load_event_with_stats_prefetch(event_id)
+    if event is None:
+        raise_validation(Code.Event.NOT_FOUND, status_code=404)
+    # Actor already has the fresh event in this response, so exclude them.
+    broadcast_capacity_change(event_id, exclude_user_ids={str(request.auth.pk)})
     return Status(200, _event_out(event, request.auth))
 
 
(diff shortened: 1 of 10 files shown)


backend/users/models.py
```diff
@@ -174,6 +174,9 @@ def create_for_user(
 
 # Number of days a non-member RSVP-management link stays valid from issuance.
 NON_MEMBER_RSVP_TOKEN_TTL_DAYS = 90
+# Absolute cap on a single token's life, even with repeated extension: a link
```
leahpeker: reduce comment to 1 line

backend/users/models.py
```diff
@@ -241,17 +244,21 @@ def issue_or_extend(cls, user: "User") -> "NonMemberRsvpToken":
         """Return a usable non-member RSVP token, reusing one when possible.
```
leahpeker, line 243: reduce this verbose descriiption to max 3 lines

backend/users/models.py
leahpeker: done — reduced to a single line in 74a5802.

backend/users/models.py
leahpeker: done — trimmed the docstring to 3 lines in 74a5802.

backend/community/_event_rsvps.py
```diff
@@ -177,11 +178,11 @@ def upsert_rsvp(request, event_id: UUID, payload: RSVPIn):
...
     )
-    event = (
-        Event.objects.select_related("created_by")
-        .prefetch_related("co_hosts", "invited_users", "rsvps__user")
-        .get(id=event_id)
-    )
+    event = _load_event_with_stats_prefetch(event_id)
+    if event is None:
+        raise_validation(Code.Event.NOT_FOUND, status_code=404)
+    # Actor already has the fresh event in this response, so exclude them.
```
leahpeker, line 184: remove line

PR #726, opened 2026-07-12
fix(events,notifications): harden update_event atomicity + follow-up cleanups (Issue 721)

### LINKED ISSUE #721: fix(events): wrap update_event field edits + status transition in a transaction (regression from #714)
## Summary

`update_event` (`backend/community/_events.py`) applies field edits **before** the status transition, but the two steps are not wrapped in a transaction. Introduced by #714 (commit `8861a71`), which reordered `_apply_field_updates()` ahead of `_handle_status_update()` to fix stale-draft date validation (Issue 627).

Because there is **no `transaction.atomic` wrapper and no `ATOMIC_REQUESTS`** in `backend/config/settings.py`, each `event.save()` autocommits. If the status transition raises *after* the field edits are saved, the field edits persist even though the client receives a 4xx and reasonably assumes the whole PATCH was rejected.

## Reproduction

`PATCH /api/community/events/{id}/` with `{ "title": "new title", "status": "active" }` on a DRAFT whose `start_datetime` is in the past:

1. `_apply_field_updates()` saves `title="new title"` and emits an `event_updated` audit log.
2. `_handle_status_update()` → `_publish_draft()` raises `START_DATETIME_MUST_BE_FUTURE` (400).
3. Client gets a 400 and assumes nothing saved — but the title change is committed.

Same partial-write occurs for:
- A co-host sending `status:"active"` on a cancelled event (403 uncancel-denied) + a field edit.
- Any `INVALID_STATUS_TRANSITION` paired with a field edit.

## Secondary effects (same root cause)

- **Cancellation notice uses the new title:** `{title:"Renamed", status:"cancelled", notify_attendees:true}` saves the title first, so `create_event_cancellation_notifications` notifies attendees with a name they never saw.
- **Spurious audit log on delete:** `{title:"x", status:"deleted"}` now saves the field edit + logs `event_updated` on a row it immediately soft-deletes.

## Fix

Wrap the field-edit + transition sequence in `transaction.atomic()` — keep the new ordering (it correctly fixes Issue 627), just make it atomic so a failed transition rolls back the field edits:

```python
with transaction.atomic():
    _apply_field_updates(request, event, event_id, updates)
    if new_status is not None:
        early = _handle_status_update(request, event, new_status, notify_attendees)
        if early is not None:
            return early
```

## Tests to add

- PATCH combining a field edit + a failing status transition asserts the field edit did **not** persist (re-fetch from DB).
- The existing `test_publish_stale_draft_with_corrected_date_in_one_patch` (added in #714) should still pass.

leahpeker:
Closes #721.

Follow-up to the code-review findings on #714 (which was merged before the review landed). Addresses all 5 findings.

## The regression (finding 1 — the reason for this PR)

#714 reordered `update_event` so field edits run **before** the status transition (to fix stale-draft date validation, Issue 627), but the two steps weren't atomic and there's no `ATOMIC_REQUESTS`. A PATCH combining a field edit with a status transition that then fails committed the field edit while returning a 4xx — the client saw an error but the edit persisted.

**Fix:** wrap both steps in `transaction.atomic()`, keeping the new ordering. A rejected transition now rolls back the field edits.

Regression test `test_field_edit_rolls_back_when_transition_rejects` verifies a `{title, status:"draft"}` PATCH on an active event (invalid transition) leaves the title unchanged. Confirmed it fails without the wrapper and passes with it.

## Other findings

- **Finding 3 — notifications race:** `useMarkAllNotificationsRead.onMutate` now `await qc.cancelQueries(...)` before the optimistic write, matching the convention in `eventComments.ts` / `eventPolls.ts`, so a late refetch can't clobber the optimistic write or stale the `onError` rollback snapshot.
- **Finding 5 — silent-failure backstop:** added a `MutationCache` `onError` to `queryClient` that `reportError`s every mutation failure to the backend. It's **report-only** (no toast) so components keep owning user-facing messaging and there's no double-notify — this kills the silent-failure class at the root instead of one call site at a time.
- **Finding 4 — duplication:** extracted `removeMyRsvp()` in `RsvpSection`, shared by `apply()`'s remove branch and leave-waitlist (was a byte-for-byte copy).
- **Finding 6 — release script:** `set_release_version` now scopes the version rewrite to the `[project]` table rather than matching by value, so a `version =` line in another table can't spuriously block a release. Updated the corresponding test to assert the correct new behavior (other tables left untouched).

## Not addressed (deliberately out of scope)

The finding-2 secondary effect (a cancellation notice using a newly-saved title within the same PATCH) is now correct-by-atomicity — if the whole PATCH succeeds, the event genuinely has the new title. No separate change needed.

## Verification

- Backend: ruff ✓, format ✓, 83 event-related + release tests ✓ (isolated SQLite). One pre-existing `ty` warning on untouched `_events.py:77` (`dict.get` enum key) — not introduced here.
- Frontend: tsc ✓, eslint ✓, prettier ✓, 31 affected tests ✓.


leahpeker:
superseded. of the 5 findings here, 3 landed on main independently over the past week (update_event atomicity via transaction.atomic, RsvpSection removeMyRsvp dedup, set_release_version [project] scoping) and the notifications onError snapshot landed partially. the two residuals — awaiting cancelQueries before the optimistic mark-all, and the report-only MutationCache backstop — are carried in #951. closing this in favor of that.

PR #748, opened 2026-07-13
fix(join-requests): clear logged-in users from approved list (Issue 725)

### LINKED ISSUE #725: A single person who logged in 2 months ago is still showing in the approved lost
## Context

feedback from `/join-requests`: a single person who logged in ~2 months ago is still showing in the **approved** list. once someone has logged in they should have transitioned out of the approved/pending state into full membership, so a logged-in user lingering in approved looks like a state-transition bug (or a data anomaly on one record).

## What needs to happen

investigate why a user who has already logged in still appears in the approved list, and correct both the underlying transition logic (if it's a bug) and the affected record.

## Acceptance criteria

- [ ] root cause identified: is this a bad single record, or a systematic gap in the approved → member transition on first login?
- [ ] logged-in users no longer appear in the approved list
- [ ] the specific affected record is corrected

## Open Questions

- which user is it (the report doesn't name them)? identify via the join-requests / approved data.
- likely related to the never-logged-in / archive semantics in #722 — may share a fix.

---

## Original

> In join requests

**Route:** `/join-requests`
_Submitted by user id: `a7de2dd4-3f09-4403-8ea1-65ee9479f092`_ (iPhone Safari)

leahpeker:
## Summary

Closes #725

A user who logged in ~2 months ago still lingered in the admin `/join-requests` **approved** list. Root cause: the approved-list exclusion in `list_join_requests` only dropped users who **completed the onboarding form** (which sets `onboarded_at` / clears `needs_onboarding`) — login itself recorded nothing. So anyone who authenticated via a magic link but never finished onboarding stayed in *approved* forever, regardless of how long ago they logged in. Per issues #725 and #722, the approved list is meant to hold people who have **not yet logged in**.

- record `last_login` on **first** login and magic-login (`stamp_first_login`) — first-only so the grace timer anchors to first login rather than sliding forward on every login
- broaden the `list_join_requests` exclusion to also drop approved requests whose linked user logged in past the 3-day grace window
- expose `last_login` on `JoinRequestOut`; the admin "logged in / hasn't logged in yet" status now reflects actual login rather than onboarding-form completion
- data migration `0036` backfills `last_login` from the most recent **used** magic token to correct existing lingering records (login never stamped it historically)

## Test plan

- [x] backend: `test_login_stamps_last_login`, `test_login_does_not_refresh_existing_last_login`, `test_magic_login_stamps_last_login`
- [x] backend: `TestJoinRequestListVisibility` — logged-in-but-never-onboarded user excluded after grace, included within grace
- [x] frontend: `join.test.ts` maps `lastLogin`; full vitest suite green
- [x] `migrate users` applies `0036` cleanly; backfill keys off the latest `used=True` token
- [ ] on staging: confirm a formerly-lingering approved record clears once backfilled / after next login

### DIFF
--- a/backend/community/_join_requests.py
+++ b/backend/community/_join_requests.py
@@ -189,9 +189,9 @@ def list_join_requests(request):
         raise_validation(Code.Perm.DENIED, status_code=403, action="list_join_requests")
 
     cutoff = timezone.now() - timedelta(days=APPROVED_GRACE_DAYS)
-    expired_phones = User.objects.filter(
-        needs_onboarding=False, onboarded_at__lt=cutoff
-    ).values_list("phone_number", flat=True)
+    expired_phones = User.objects.filter(onboarded_at__lt=cutoff).values_list(
+        "phone_number", flat=True
+    )
     # Legacy users onboarded before onboarded_at existed have it as null;
     # treat them as already-expired so they don't linger in the list forever.
     legacy_onboarded_phones = User.objects.filter(
--- a/backend/tests/test_join_request_list_visibility.py
+++ b/backend/tests/test_join_request_list_visibility.py
@@ -1,0 +1,52 @@
+"""Tests for onboarding-driven visibility of approved join requests in the list."""
+
+from datetime import timedelta
+
+import pytest
+from community.models import JoinRequestStatus
+from django.utils import timezone
+from users.models import User
+
+
+@pytest.mark.django_db
+class TestJoinRequestListVisibility:
+    def test_list_excludes_reapproved_onboarded_user_past_grace(
+        self, api_client, vettor_headers, sample_join_request
+    ):
+        api_client.patch(
+            f"/api/community/join-requests/{sample_join_request.id}/",
+            {"status": JoinRequestStatus.APPROVED},
+            content_type="application/json",
+            **vettor_headers,
+        )
+        # onboarded_at is set once and never cleared; a re-approval flips
+        # needs_onboarding back to True, leaving the contradictory state that
+        # used to keep the request in the approved list forever.
+        user = User.objects.get(phone_number=sample_join_request.phone_number)
+        user.onboarded_at = timezone.now() - timedelta(days=60)
+        user.needs_onboarding = True
+        user.save(update_fields=["onboarded_at", "needs_onboarding"])
+
+        response = api_client.get("/api/community/join-requests/", **vettor_headers)
+        assert response.status_code == 200
+        ids = [r["id"] for r in response.json()]
+        assert str(sample_join_request.id) not in ids
+
+    def test_list_includes_onboarded_user_within_grace(
+        self, api_client, vettor_headers, sample_join_request
+    ):
+        api_client.patch(
+            f"/api/community/join-requests/{sample_join_request.id}/",
+            {"status": JoinRequestStatus.APPROVED},
+            content_type="application/json",
+            **vettor_headers,
+        )
+        user = User.objects.get(phone_number=sample_join_request.phone_number)
+        user.onboarded_at = timezone.now() - timedelta(days=1)
+        user.needs_onboarding = False
+        user.save(update_fields=["onboarded_at", "needs_onboarding"])
+
+        response = api_client.get("/api/community/join-requests/", **vettor_headers)
+        assert response.status_code == 200
+        ids = [r["id"] for r in response.json()]
+        assert str(sample_join_request.id) in ids


backend/community/_join_requests.py
```diff
@@ -137,10 +139,17 @@ def list_join_requests(request):
     legacy_onboarded_phones = User.objects.filter(
         needs_onboarding=False, onboarded_at__isnull=True
     ).values_list("phone_number", flat=True)
+    # Anyone who logged in past the grace window has transitioned to a member even
```
leahpeker: remove comment

backend/tests/test_auth.py
```diff
@@ -1,5 +1,7 @@
 """Tests for auth endpoints: login, refresh, me, change-password, onboarding."""
```
leahpeker, line 1: remove comment

backend/tests/test_join_request_list_visibility.py
```diff
@@ -0,0 +1,51 @@
+"""Tests for login-driven visibility of approved join requests in the list."""
```
leahpeker: remvoe comment

backend/users/_helpers.py
```diff
@@ -10,11 +10,22 @@
 import phonenumbers
 from community._shared import validate_display_name
 from community._validation import Code, raise_validation
+from django.utils import timezone
 
 from users.models import MagicLoginToken, User
 from users.roles import Role
 
 
+def stamp_first_login(user: User) -> None:
```
leahpeker: i dont undetstand why we would need this

also the user in question has a logged in at time on their user so how does this even help

leahpeker: good questions — two separate things are going on here.

**why we need `stamp_first_login`:** our JWT login flow (`/api/auth/login/` and `/api/auth/magic-login/`) never calls django's `django.contrib.auth.login()`, so django's built-in `update_last_login` signal never fires. nothing in the API path was recording `last_login` at all. that's the original bug: a user could magic-login and never onboard, and since `last_login` stayed null AND `onboarded_at` stayed null, they'd sit in the approved list forever. `stamp_first_login` is what actually records login time for the API flows so the new `last_login__lt=cutoff` exclusion has something to key off of going forward.

**why the reported user already has a `last_login`:** two possible sources — (1) the django **admin** login *does* call `login()` and stamps it, so if they ever hit `/admin/` they'd have one; and (2) migration `0037` backfills `last_login` from their most recent used magic token. so for *this specific user*, the fix that clears them is the exclusion + the backfill, not `stamp_first_login`.

so to be precise about "how does this even help": `stamp_first_login` doesn't fix the existing user — the migration + exclusion do. `stamp_first_login` prevents the bug from **recurring** for every future never-onboarded user, since otherwise API logins would keep failing to record login time and new users would linger exactly the same way.

if you'd rather not carry the helper, the alternative is wiring django's `update_last_login` into the JWT flow — but that refreshes on *every* login, which would let a never-onboarded user slide their own grace window forward indefinitely (that's the case `test_login_does_not_refresh_existing_last_login` guards). happy to go that route if you prefer, just with the first-login-only guard kept.

backend/community/_join_requests.py
leahpeker: done in ed8f068

backend/tests/test_auth.py
leahpeker: done in ed8f068

backend/tests/test_join_request_list_visibility.py
leahpeker: done in ed8f068

leahpeker:
you were right to push on this — digging into the actual record (`onboarded_at = 2026-05-01`, `needs_onboarding = True`, phone already E.164) showed the `last_login` approach was solving the wrong problem. writing up the real root cause and the revised fix.

**real root cause:** onboarding sets `onboarded_at` **and** `needs_onboarding=False` together and never clears `onboarded_at`. but re-approval / promotion (`_join_request_approval.py`) flips `needs_onboarding` back to `True` without touching `onboarded_at`. that leaves the contradictory state this user is in: onboarded 2.5 months ago, but `needs_onboarding=True` again.

both existing exclusions in `list_join_requests` gate on `needs_onboarding=False`:
```python
expired_phones          = User.filter(needs_onboarding=False, onboarded_at__lt=cutoff)
legacy_onboarded_phones = User.filter(needs_onboarding=False, onboarded_at__isnull=True)
```
so a re-approved-but-already-onboarded user fails both → never excluded → lingers forever. the `last_login` filter only "worked" because it was the one condition that *didn't* check `needs_onboarding` — i.e. it masked the bug rather than fixing it.

**revised fix:** drop the `needs_onboarding=False` gate from the onboarded-past-grace exclusion — someone who onboarded more than the grace window ago has transitioned, even if a re-approval reset `needs_onboarding`:
```python
expired_phones = User.objects.filter(onboarded_at__lt=cutoff).values_list("phone_number", flat=True)
```
(the legacy null-`onboarded_at` filter keeps its `needs_onboarding=False` gate — otherwise brand-new approved users who haven't onboarded would wrongly drop off the list.)

**removed all the `last_login` machinery** as unnecessary: `stamp_first_login` + its call sites, the freeze logic, the three `test_login_*`/magic tests, the `0037` backfill migration, and the `JoinRequestOut.last_login` field + frontend `lastLogin` plumbing (reverted `LoginStatus` back to `onboardedAt`). net −97 lines.

test coverage now targets the actual scenario: `test_list_excludes_reapproved_onboarded_user_past_grace` sets `onboarded_at` 60 days ago + `needs_onboarding=True` and asserts exclusion (this would fail on the old `needs_onboarding=False`-gated filter).

green locally: backend tests (52), ruff, ty, tsc, vitest (join + JoinRequestsScreen).

one thing i left alone deliberately: this fixes the *listing*, not the *re-approval* semantics — re-approval still produces the `onboarded_at` set + `needs_onboarding=True` combination. if forcing an already-onboarded user back through onboarding on re-approval is itself wrong, that's a separate change to `_join_request_approval.py`; happy to file/fix that too if you want.

backend/users/_helpers.py
leahpeker: update: your instinct was right — dug into the actual record and the whole last_login approach was masking the real bug (a re-approval leaving onboarded_at set + needs_onboarding=True, which both list exclusions skipped). stamp_first_login is gone; see the top-level comment for the revised fix. disregard my reasoning above.

PR #749, opened 2026-07-13
docs: 717 - explore whether to consolidate seeders

### LINKED ISSUE #717: Investigate whether local and staging seeders should be consolidated
## Context

The seed layer under `backend/community/management/commands/` currently has **two pairs** of files:

| Command | Data module | Purpose |
|---|---|---|
| `seed.py` (229 LOC) | `_seed_data.py` (311 LOC) | local dev seed (`make seed`) |
| `seed_staging.py` (223 LOC) | `_seed_staging_data.py` (188 LOC) | staging seed (`seed_staging`, staging-gated) |

Within each pair, the command imports its `_*_data.py` module — that split (orchestration logic vs static data + pure helpers) is **intentional** and keeps each file under the ~300-line limit, so "do we need both files in a pair" is not the real question.

The question raised in conversation is whether we need **both pairs** — i.e. is there enough overlap between the local (`seed` / `_seed_data`) and staging (`seed_staging` / `_seed_staging_data`) seeders that they should be consolidated behind one command (e.g. `seed --target=local|staging`, or a shared data module both import), rather than maintained as two parallel code paths that can drift.

Note: `_seed_staging_data.py` was just extended in #711 (non-member RSVP seeding), and there's a design spec at `docs/superpowers/specs/2026-07-09-seed-staging-design.md` documenting the staging seeder's intent — read it before proposing changes.

## What needs to happen

Investigate and recommend (spike, not necessarily a rewrite):

- Compare `_seed_data.py` and `_seed_staging_data.py`: how much is genuinely duplicated (event/user/role fixtures, helpers) vs legitimately environment-specific (staging's per-permission role users, condition-matrix users, non-member RSVP band, staging gating)?
- Same for `seed.py` vs `seed_staging.py` orchestration.
- Decide: keep two independent pairs, extract a shared data/helper module both import, or unify into one command with a target flag.
- If they should stay separate, document *why* (so this doesn't get re-raised) — likely: staging needs deterministic per-permission fixtures + env gating that local dev doesn't want.

## Acceptance criteria

- [ ] A written recommendation: consolidate vs keep-separate, with the overlap/divergence evidence that drove it.
- [ ] If consolidating: a concrete plan (shared module or unified command) that preserves staging's env gate and idempotency.
- [ ] If keeping separate: a short rationale captured in code/docs so the duplication is understood as deliberate.

leahpeker:
## Overview

Investigation for issue #717: should the local (`seed` / `_seed_data`) and staging (`seed_staging` / `_seed_staging_data`) seeders be consolidated?

**Recommendation: keep the two pairs separate.** The seeders share only a thin skeleton (`_seed_events` + the generic `get_or_create` user idiom); their data domains, safety posture, tests, and invocation paths are essentially disjoint:

- **Disjoint data** — local seeds join-form questions, RSVPs, join requests, and content singletons; staging seeds enum-driven per-permission roles/users, an 8-way condition matrix, and a non-member RSVP band. No shared users (disjoint phone blocks), roles, or content.
- **Different safety posture** — staging is env-gated (`RAILWAY_ENVIRONMENT_NAME`), transaction-wrapped, and resettable; local is none of these.
- **Disjoint consumers** — separate test suites and invocation sites (`make seed`/dev-db scripts vs. an on-demand Railway one-off).
- **Documented intent** — the staging seeder exists separately "so `make seed` never emits demo accounts."

Consolidating into one `--target` command would add a fork branch and endanger that guarantee for no real simplification. The spec recommends documenting the split, optionally deduping just `_seed_events`, and fixing an incidental affected-tests gap (`test_seed_staging.py` missing from `scripts/list_affected_tests.py`).

Full findings, evidence, and options: `docs/2026-07-13-consolidate-seeders-explore.md`.

Note: the spec is placed at the tracked `docs/` root rather than the skill's default `docs/superpowers/specs/` because that path was deliberately gitignored in `006decd`. See the spec's Open Questions.

Closes #717

## Test plan

- [ ] N/A — investigation only; adds one findings-spec markdown file, no code changes.

### DIFF
--- a/docs/2026-07-13-consolidate-seeders-explore.md
+++ b/docs/2026-07-13-consolidate-seeders-explore.md
@@ -1,0 +1,181 @@
+# Explore: investigate whether local and staging seeders should be consolidated (#717) — Findings
+
+**Date:** 2026-07-13
+**Issue:** https://github.com/ProteinDeficientsAnonymous/pda/issues/717
+**Branch:** auto-717-explore-seeder-consolidation
+
+## The ask
+
+The seed layer under `backend/community/management/commands/` has two command/data pairs:
+
+| Command | Data module | Purpose |
+|---|---|---|
+| `seed.py` | `_seed_data.py` | local dev seed (`make seed`) |
+| `seed_staging.py` | `_seed_staging_data.py` | staging seed (`seed_staging`, env-gated) |
+
+The within-pair split (orchestration vs static data) is intentional and stays. The
+real question is whether the **two pairs** overlap enough to consolidate behind one
+command (e.g. `seed --target=local|staging`) or a shared data/helper module — rather
+than maintaining two parallel code paths that can drift.
+
+## What we found
+
+**The two seeders share a skeleton but seed almost entirely disjoint data for
+different audiences.**
+
+### Structural overlap is real but shallow
+
+The one genuinely duplicated piece of orchestration is `_seed_events`. The two
+implementations are near-identical: both walk a `*_EVENTS` list, compute
+`start`/`end` from `delta_days`/`duration_hours` off `timezone.now()`, and
+`get_or_create` by `title` with the same core defaults
+(`seed.py:128-152` vs `seed_staging.py:154-173`). Local additionally sets
+`rsvp_enabled`, `allow_plus_ones`, `max_attendees`; staging omits those. Beyond
+that, both share only the generic `get_or_create` user/role idiom and a shared test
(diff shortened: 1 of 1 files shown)


leahpeker:
Closing — the two required follow-ups from this spec (rationale comment + affected-tests fix) landed in #936. The optional _seed_events dedup was skipped per the spec's own recommendation (nice-to-have, not required).

PR #752, opened 2026-07-13
feat(users): strip non-member roles when a user is paused (Issue 708)

### LINKED ISSUE #708: strip non-member roles when a user is paused
## Context

Surfaced while refactoring role-member counting (PR #703). Paused members currently keep all their roles: `_apply_user_patch` (`backend/users/_management.py:298-299`) only sets `is_paused = True` and leaves the role M2M untouched. Because of that, paused members still appear when filtering the member directory by role (`MembersTab.tsx`), and still hold whatever elevated permissions their roles grant at the data level (auth is blocked at the gate, but the role assignments persist).

Decision: **pausing a user should strip all of their roles except the default `member` role.** This makes a paused user hold no elevated role, so they naturally drop out of role filters/counts and lose role-granted permissions — no separate view-filtering needed.

Unpause behavior (decided): roles stay stripped. Unpausing only restores login; an admin re-grants any roles manually. No snapshot of the pre-pause role set is stored.

## What needs to happen

- On the pause transition (`is_paused` false → true) in `_apply_user_patch` / the pause flow (`backend/users/_management.py`), remove every role from the user except the default `member` role.
- Only act on the false → true transition (don't re-strip on unrelated PATCHes to an already-paused user).
- Audit-log the role removal (reuse/extend the existing `user_paused` audit entry with the removed role ids).
- Unpause: no role changes (leave as-is).
- Admins already can't be paused (`_validate_pause_change`, `_management.py:284-285`), so the admin role / last-admin guard is unaffected.

## Acceptance criteria

- [ ] Pausing a member with non-member roles leaves them holding only the default `member` role
- [ ] Pausing a member who only has `member` is a no-op for roles
- [ ] Unpausing does not restore roles
- [ ] Paused members no longer appear under non-member role groups in the directory filter
- [ ] The role removal is captured in the audit log
- [ ] Tests cover the pause-strip, the member-only no-op, and the unpause-no-restore cases

leahpeker:
## Summary

- On the `is_paused` false→true transition in `update_user`, strip every role from the user except the default `member` role, so a paused member drops out of role filters/counts and loses role-granted permissions (auth is already blocked at the gate; this removes the persistent data-level grants).
- Only fires on the pause transition — a PATCH to an already-paused user does not re-strip, and unpausing does not restore roles (an admin re-grants manually; no snapshot is stored).
- The removed role ids are captured in the existing `user_paused` audit entry as `details.removed_role_ids`.
- Admins already can't be paused (`_validate_pause_change`), so the last-admin/admin-role guard is unaffected.
- Extracted the post-save audit logic into `_audit_user_update` to keep `update_user` within the cognitive-complexity limit.

Closes #708

## Test plan

- [x] `test_pause_strips_non_member_roles` — pausing a member with an extra role leaves only `member`; audit log records the removed role id
- [x] `test_pause_member_only_is_noop_for_roles` — pausing a member-only user leaves `member`; audit log records an empty `removed_role_ids`
- [x] `test_unpause_does_not_restore_roles` — unpausing does not restore stripped roles
- [x] Full `test_users_admin_extended.py` suite passes (37/37) on SQLite; lint, typecheck, complexity, and schema-code checks green

### DIFF
--- a/backend/users/_management.py
+++ b/backend/users/_management.py
@@ -21,6 +21,7 @@ from users._helpers import (
     _is_admin,
     _is_last_admin,
     _resolve_name_fields,
+    _strip_non_member_roles,
     _validate_admin_role_change,
     _validate_member_role_required,
     _validate_phone,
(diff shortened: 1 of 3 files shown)


backend/users/_management.py
```diff
@@ -297,26 +298,41 @@ def update_user(request, user_id: str, payload: UserPatchIn):
...
+        details = {"removed_role_ids": _strip_non_member_roles(user)} if payload.is_paused else None
+        audit_log(
+            logging.WARNING, action, request, target_type="user", target_id=user_id, details=details
+        )
+        return
 
-    return Status(200, UserOut.from_user(user))
+    changed = [
+        f
+        for f in ("phone_number", "display_name", "email")
```
leahpeker: we are in the process of removing display_name from users

leahpeker: good catch — the PATCH actually mutates first_name/last_name, so display_name in the audit fields list was wrong. fixed in c164c4f: the changed-fields list now reports first_name/last_name, and a legacy display_name-only patch maps to both. display_name is no longer referenced in this audit path.

PR #757, opened 2026-07-13
docs: 699 - explore findings for slow backend tests

### LINKED ISSUE #699: Investigate and speed up slow backend tests
## Context

The backend pytest suite is slow enough to be a drag on local iteration and CI. A full run measured **243s (~4m) for 1186 tests across 8 xdist workers** (`uv run pytest backend/tests/ --durations=25`).

The slowest offenders cluster in a few files, and the cost is concentrated in a handful of patterns:

**1. `test_seed_staging.py` — full seed command re-run per test (biggest lever)**
The three slowest tests in the whole suite live here, and the file calls the full `seed_staging` management command ~11 times:
- `test_seed_staging_creates_one_role_per_permission` — **18.0s**
- `test_seed_staging_perm_users_hold_only_their_role_and_are_onboarded` — **17.4s**
- `test_seed_staging_reset_removes_only_scoped_rows` — **11.4s**
- plus `creates_eight_member_only_condition_users` (8.9s), `events_span_past_current_future` (6.4s), `is_idempotent` (5.9s), `perm_and_condition_users_are_disjoint` (5.8s)

Each `call_command("seed_staging")` rebuilds the entire staging dataset. A shared module/class-scoped fixture that seeds once and asserts against the result (instead of re-seeding per assertion group) would likely reclaim most of ~90s.

**2. Expensive per-test fixture setup (setup time, not call time)**
Several suites show 5-10s in **setup** — fixtures rebuilt for every test rather than shared:
- `test_event_capacity.py::TestWaitlistPromotion` — setup 6-10s per test (`test_promote_on_status_change` 10.0s, `test_fifo_order` 9.2s, `test_promote_on_delete` 6.7s, `test_promote_multiple_spots` 6.1s)
- `test_event_stats.py::TestStatsEndpoint` / `TestAttendanceCounts` / `TestSetAttendance` / `TestNoResponseCount` — setup ~5-6s across ~9 tests
- `test_event_cohost_invites.py::TestEventOutVisibility` — setup 7.3s
- `test_event_email_blast.py` — setup 5.8s
- `test_auth_update_me.py` — setup 5.6s

**3. Inline bulk creation in a single test**
- `users/test_users_crud.py::TestSearchUsers::test_search_limits_to_ten_results` — **8.6s** call, likely creating many users inline where a bulk-create or smaller dataset would do.

## What needs to happen

Investigate the slow backend tests and identify concrete, low-risk speedups. Focus first on the highest-leverage items above:

- Determine whether `test_seed_staging.py` can seed once via a shared fixture (module/class scope) instead of re-running `call_command("seed_staging")` per test, without weakening coverage.
- Audit the heavy fixtures in `test_event_capacity` (`TestWaitlistPromotion`), `test_event_stats`, `test_event_cohost_invites`, `test_event_email_blast`, and `test_auth_update_me` — are they doing more DB work than each test needs? Can setup be shared or trimmed?
- Check `test_search_limits_to_ten_results` for cheaper data setup (`bulk_create`, fewer rows).
- Confirm whether these are genuinely re-run per test vs. an xdist worker-startup artifact.
- Quantify the win: re-measure with `--durations` after changes and report the before/after wall-clock.

Keep behavior/coverage identical — this is a test-performance investigation, not a rewrite.

## Acceptance criteria

- [ ] Root cause documented for the top offenders (`test_seed_staging`, `TestWaitlistPromotion` setup, `test_event_stats` setup, `test_search_limits_to_ten_results`)
- [ ] Concrete optimization proposed (and ideally applied) for at least the `seed_staging` re-run pattern
- [ ] Measurable reduction in total suite wall-clock, reported as before/after (`--durations` output)
- [ ] Full suite still green (`make agent-test`) with no loss of coverage

leahpeker:
## Overview

Investigation (spike) for issue #699 — the backend pytest suite is slow (~243s / 1186 tests / 8 workers). This PR adds a findings spec documenting the root cause and a ranked set of low-risk speedups.

**Root cause:** tests run under `config.settings` with **no `PASSWORD_HASHERS` override**, so every `create_user(...)` pays a full default PBKDF2-SHA256 hash (~600k iterations). The flagged offenders (`test_seed_staging`, `TestWaitlistPromotion`/`test_event_stats`/etc. setup, `test_search_limits_to_ten_results`) are exactly the suites that create many users per test.

**Empirically measured on this worktree (SQLite, single worker):**
- `test_auth_update_me.py`: 4.96s → 1.77s (~64%) with a fast MD5 hasher
- `test_seed_staging.py` 2 slowest tests: 16.58s → 1.40s (~92%)

**Recommendation:** ship a fast test password hasher first (one small change, identical coverage, biggest win), then add `--no-migrations` to the pytest/Makefile addopts, and only refactor fixture scope for any residual hot suites after re-measuring. Full detail, citations, options, and open questions in the spec.

This is an `explore`-labeled investigation — it recommends but does not implement the code change.

Spec: `docs/699-speed-up-slow-backend-tests-explore.md`

Closes #699

### DIFF
--- a/docs/699-speed-up-slow-backend-tests-explore.md
+++ b/docs/699-speed-up-slow-backend-tests-explore.md
@@ -1,0 +1,86 @@
+# Explore: Investigate and speed up slow backend tests (#699) — Findings
+
+**Date:** 2026-07-13
+**Issue:** https://github.com/ProteinDeficientsAnonymous/pda/issues/699
+**Branch / PR:** `auto-699-explore-slow-backend-tests` (draft PR linked below)
+
+## The ask
+
+The backend pytest suite is slow (measured **243s / ~4m for 1186 tests across 8 xdist workers**). The issue lists the top offenders by file:
+
+- `test_seed_staging.py` — the three slowest tests in the whole suite (18.0s / 17.4s / 11.4s); the file re-runs the full `seed_staging` command ~11 times.
+- Heavy per-test **fixture setup** (5–10s) in `test_event_capacity.py::TestWaitlistPromotion`, `test_event_stats.py`, `test_event_cohost_invites.py`, `test_event_email_blast.py`, `test_auth_update_me.py`.
+- `test_users_crud.py::TestSearchUsers::test_search_limits_to_ten_results` — 8.6s inline bulk creation.
+
+Investigate the root cause of the top offenders, propose concrete low-risk speedups, and quantify the win — keeping behavior/coverage identical.
+
+## What we found
+
+**The single root cause behind nearly every offender is that tests use Django's default password hasher.** There is **no `PASSWORD_HASHERS` override anywhere** in the repo (confirmed: `grep -rn PASSWORD_HASHERS backend/` returns nothing). `config/settings.py:81` defines only `AUTH_PASSWORD_VALIDATORS`; there is no test settings module and no pytest branch, so tests run under `config.settings` with the default **PBKDF2-SHA256 hasher (~600k iterations)**.
+
+`UserManager.create_user` (`backend/users/models.py:61-66`) calls `user.set_password(password)` unconditionally, so **every `create_user(...)` in every fixture and test pays a full PBKDF2 hash**. The flagged suites are exactly the ones that create many users per test, and the "slow setup" in `--durations` is that hashing.
+
+**Empirically verified on this worktree** (SQLite dev.db, single worker, PBKDF2 vs. injected `MD5PasswordHasher`):
+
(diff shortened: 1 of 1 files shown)


leahpeker:
Closing — Option A (fast MD5 test hasher) already shipped in `backend/tests/conftest.py`. Option C investigated and mostly not worth doing (suite is 13.67s post-A, nothing hot); the one safe win landed in #934. Option B (`--no-migrations`) investigated and found unsafe as-is — breaks 48 tests that rely on data-migration-seeded `Role` rows, since `--no-migrations` skips migration code entirely. Not pursuing further given the marginal remaining win.

PR #759, opened 2026-07-13
feat(events): default event end to start + 2h when omitted (Issue 697)

### LINKED ISSUE #697: ends is optional
## Context

feedback from `/events/add`: the event **end** time currently reads as required (or its behavior is unclear). the reporter suggests making it optional with a sensible default so people know what to expect.

## What needs to happen

make the event end time **optional**, defaulting to **2 hours** after the start when left blank, and surface that default in the UI so users understand it.

## Acceptance criteria

- [ ] end time is not required on the add-event form
- [ ] when omitted, end defaults to start + 2 hours
- [ ] the 2-hour default is communicated in the UI (helper text / placeholder), lowercase per project convention
- [ ] backend accepts a missing end and applies the same default

## Open Questions

- should the default be applied on the backend (canonical) or only computed in the UI? recommend backend so all clients agree.

---

## Original

> maybe we can add optional - defaults 2hrs
> so that people know

**Route:** `/events/add`

leahpeker:
## Summary
- event **end** time is now optional on the add-event form; when left blank the backend fills it with **start + 2 hours** (canonical, so all clients agree) via a new `Event.default_end()` helper + `DEFAULT_EVENT_DURATION` constant
- the ICS calendar builder reuses `Event.default_end()` instead of its own inline `start + 2h` fallback
- the "ends" field in the form shows helper text — "defaults to 2 hours after start if left blank" — via a new optional `hint` prop on `DateTimePicker`
- the PATCH/update path is intentionally left unchanged, so an event with a start can still be made open-ended by clearing its end (existing behavior preserved)

Closes #697

## Test plan
- [ ] create an event with a start but no end → end persists as start + 2h
- [ ] the add-event form "ends" field shows the lowercase helper text
- [ ] editing an existing event and clearing its end still leaves it open-ended (null)
- [ ] the event's ICS feed entry uses start + 2h when no explicit end is set
- [ ] backend: `test_event_management.py` passes (incl. updated default-end test)

### DIFF
--- a/backend/community/_calendar.py
+++ b/backend/community/_calendar.py
@@ -174,7 +174,7 @@ def _build_vevent(event, request: HttpRequest, is_authed: bool):
         vevent.add("dtstart", event.start_datetime)
         vevent.add(
             "dtend",
-            event.end_datetime or event.start_datetime + timedelta(hours=2),
+            Event.default_end(event.start_datetime, event.end_datetime),
         )
     vevent.add("summary", event.title)
     target_url = request.build_absolute_uri(f"/events/{event.id}")
--- a/backend/community/_events.py
+++ b/backend/community/_events.py
@@ -329,7 +329,7 @@ def create_event(request, payload: EventIn):
         title=payload.title,
         description=payload.description,
         start_datetime=payload.start_datetime,
-        end_datetime=payload.end_datetime,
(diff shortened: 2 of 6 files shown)


leahpeker:
closing this — the approach here wasn't well thought out. reverting to reconsider the design for defaulting event end times.

PR #761, opened 2026-07-13
feat(events): show visibility as locked read-out for official/club (Issue 695)

### LINKED ISSUE #695: when tagging club or official
## Context

feedback from `/events/add`: when an event is tagged **official** or **club** PDA, its visibility is forced to public. there is explanatory text saying so, but it doesn't stand out — users miss it.

## What needs to happen

add a clear **visual indicator** that the visibility selection is locked (not just body text). e.g. show the visibility dropdown but disable/grey out the non-public options, or a lock affordance, so it's obvious the choice is fixed to public.

## Acceptance criteria

- [ ] when official/club is tagged, the visibility control visibly reads as locked to public (disabled/greyed non-public options or equivalent)
- [ ] the locked state is discoverable without reading the explanatory sentence
- [ ] lowercase user-facing text per project convention

## Open Questions

- exact treatment (greyed dropdown vs lock icon vs inline badge) — a design call; reporter suggested greying out the other options.

---

## Original

> we have the text that official and club pda events are always public but we should have a visual indicator that it's locked, maybe like open the dropdown but gray out the other options? idk something tht makes it clearer bc the text doesnt stand out that much

**Route:** `/events/add`

leahpeker:
## Summary

Closes #695

- on `/events/add`, when an event is tagged **official** or **club** pda, visibility is forced to public. previously the dropdown was just `disabled`, which greyed uniformly and didn't clearly read as *locked* — reporters said the explanatory text "doesn't stand out".
- replace the disabled dropdown (when the type is locked) with a **locked read-out**: a lock icon, `public` as the active value, `members only` / `invite only` struck through and greyed, and a `locked` badge. the constraint is now discoverable without reading the helper sentence.
- accessible + color-blind-safe: `role="group"` labelled "who can see it — locked to public", `aria-disabled` + `sr-only` cues on the excluded options, and meaning conveyed by icon / strikethrough / text badge rather than color alone.

## Test plan

- [ ] new `EventFormDetails.test.tsx` covers: dropdown shown when unlocked; locked read-out (no combobox, `locked` badge, group label) when type-locked; non-public options greyed + `aria-disabled`.
- [ ] `make agent-frontend-typecheck`, `make agent-frontend-lint`, and full `make agent-frontend-test` (732 passed) all green.
- [ ] manual: on `/events/add`, toggle "make it an official pda event" (or club) and confirm the visibility control switches to the locked read-out.

### DIFF
--- a/frontend/src/screens/events/form/EventFormDetails.tsx
+++ b/frontend/src/screens/events/form/EventFormDetails.tsx
@@ -91,10 +91,7 @@ function LockIcon({ className }: { className?: string }) {
   );
 }
 
-// When the event type forces public, we replace the dropdown with a locked
-// read-out so the constraint reads visually (lock icon + badge), not just as
-// helper text. The greyed non-public options are shown struck to make the
-// forced choice obvious.
+// Forced-public event types show a locked read-out instead of the dropdown so the constraint reads visually.
 function LockedVisibility({ helper }: { helper: string }) {
   return (
     <div
--- a/frontend/src/screens/events/form/EventFormDetails.tsx
+++ b/frontend/src/screens/events/form/EventFormDetails.tsx
@@ -31,8 +31,8 @@ interface Props {
   values: EventFormValues;
   onChange: (patch: Partial<EventFormValues>) => void;
   errors: Partial<Record<keyof EventFormValues, string>>;
-  // Locked when the event is a public-only type (official) — visibility is
-  // forced to public and the select is disabled.
+  // Locked when the event is a public-only type (official/club) — visibility is
+  // forced to public and shown as a locked read-out instead of the dropdown.
   typeLocked: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


frontend/src/screens/events/form/EventFormDetails.tsx
```diff
@@ -54,18 +54,73 @@ export function EventFormDetails({ values, onChange, errors, typeLocked }: Props
...
+      strokeLinecap="round"
+      strokeLinejoin="round"
+    >
+      <rect x="4.5" y="9" width="11" height="7.5" rx="1.5" />
+      <path d="M7 9V6.5a3 3 0 0 1 6 0V9" />
+    </svg>
+  );
+}
+
+// When the event type forces public, we replace the dropdown with a locked
```
leahpeker: reduce comment to 1 line

leahpeker: done — reduced to a single line in 7e3bdab

PR #762, opened 2026-07-13
feat(events): gate invite button on member rsvp (Issue 688)

### LINKED ISSUE #688: Gate invite button on host permission + member RSVP
## Context

On the event detail screen, the "invite guests" affordance is currently shown to any member for whom `canInvite` is true. In `frontend/src/screens/events/EventMemberSection.tsx` the gate is:

```ts
const canInvite =
  !isCancelled &&
  !event.isPast &&
  (isCoHost || event.invitePermission === InvitePermission.AllMembers);
```

This means a non-cohost member sees the invite button as soon as the host has set `invitePermission` to `all_members` — regardless of whether that member has actually RSVP'd to the event. The intended behavior is that ordinary members can only invite others once they've committed to attending.

## What needs to happen

Tighten the invite gating for non-cohost members so the invite button:

1. Only appears when the host has allowed guests to invite others (`event.invitePermission === InvitePermission.AllMembers`) — this part already works.
2. For non-cohost members, additionally only appears once that member has RSVP'd (i.e. `event.myRsvp` reflects an attending/RSVP'd state, not `null`).

Co-hosts and event managers should retain their existing ability to invite regardless of RSVP status — the RSVP requirement applies only to the `all_members` (non-cohost) path.

Confirm the exact `RsvpStatus` value(s) that should count as "has RSVP'd" (e.g. attending vs. any non-null response) while implementing.

## Acceptance criteria

- [ ] A non-cohost member who has NOT RSVP'd does not see the invite button, even when `invitePermission === all_members`.
- [ ] A non-cohost member who HAS RSVP'd sees the invite button when `invitePermission === all_members`.
- [ ] When `invitePermission === co_hosts_only`, non-cohost members never see the invite button (unchanged).
- [ ] Co-hosts / event managers still see the invite button regardless of their own RSVP status.
- [ ] Existing `EventMemberSection` tests updated / new tests added covering the RSVP'd vs. not-RSVP'd non-cohost cases.

leahpeker:
## Summary

Closes #688

- Non-cohost members now only see the "invite members" button on the event detail screen once they've RSVP'd — `attending` or `maybe` count as committed (matches `MyEventsScreen`'s "engaged" bar); `cant_go`, `waitlisted`, and no-RSVP do not. This is layered on top of the existing `invitePermission === all_members` requirement.
- Co-hosts, the creator, and event managers (`manage_events`) keep inviting regardless of their own RSVP status — the RSVP requirement applies only to the ordinary `all_members` path.
- Backend invite authorization is unchanged; its docstring now notes the RSVP gate is intentionally a frontend-only UX nudge, not enforced server-side.

## Test plan

- [x] `EventMemberSection.test.tsx` — new `#688` suite covers: non-cohost not-RSVP'd (hidden), attending/maybe (shown), cant_go (hidden), `co_hosts_only` override (hidden), and creator / event-manager / co-host bypass with no RSVP (shown).
- [x] frontend typecheck + full vitest suite (737 passed) + eslint/prettier clean
- [x] backend ruff clean; invitation tests pass on sqlite dev.db

### DIFF
--- a/backend/community/_event_invitations.py
+++ b/backend/community/_event_invitations.py
@@ -36,14 +36,8 @@ class InviteIn(BaseModel):
 
 
 def _can_invite_to_event(user, event: Event) -> bool:
-    """Authorize an invite (permission gate only).
-
-    Hosts (creator / co-host) can always invite; managers can always invite;
-    other authed members can invite when invite_permission == ALL_MEMBERS. The
-    frontend additionally hides the button for all_members until the viewer has
-    RSVP'd (Issue 688) — that's a UX nudge only and is intentionally not
-    enforced here.
-    """
+    """Permission gate only — the RSVP nudge for all_members is frontend UX,
+    intentionally not enforced here (Issue 688)."""
     if user.has_permission(PermissionKey.MANAGE_EVENTS):
         return True
     if event.created_by_id == user.pk:
--- a/backend/community/_event_invitations.py
+++ b/backend/community/_event_invitations.py
@@ -36,10 +36,13 @@ class InviteIn(BaseModel):
 
 
 def _can_invite_to_event(user, event: Event) -> bool:
-    """Mirror the frontend canInvite gate.
+    """Authorize an invite (permission gate only).
 
     Hosts (creator / co-host) can always invite; managers can always invite;
-    other authed members can invite when invite_permission == ALL_MEMBERS.
+    other authed members can invite when invite_permission == ALL_MEMBERS. The
+    frontend additionally hides the button for all_members until the viewer has
+    RSVP'd (Issue 688) — that's a UX nudge only and is intentionally not
+    enforced here.
     """
     if user.has_permission(PermissionKey.MANAGE_EVENTS):
         return True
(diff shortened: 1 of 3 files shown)


frontend/src/screens/events/EventMemberSection.tsx
```diff
@@ -38,11 +38,17 @@ export function EventMemberSection({ event }: Props) {
   const canSeeInvited = isCoHost || canManageEvents;
   const isCancelled = event.status === EventStatus.Cancelled;
   const rsvpDisabled = !event.rsvpEnabled;
+  // Ordinary members may invite others only once they've committed to going —
```
leahpeker: remvoe tis comment

backend/community/_event_invitations.py
```diff
@@ -36,10 +36,13 @@ class InviteIn(BaseModel):
 
 
 def _can_invite_to_event(user, event: Event) -> bool:
-    """Mirror the frontend canInvite gate.
+    """Authorize an invite (permission gate only).
 
     Hosts (creator / co-host) can always invite; managers can always invite;
-    other authed members can invite when invite_permission == ALL_MEMBERS.
+    other authed members can invite when invite_permission == ALL_MEMBERS. The
```
leahpeker: reduce comment to 1-2 lines

PR #769, opened 2026-07-13
docs: 541 - explore rbac cedar + view/edit granularity

### LINKED ISSUE #541: RBAC updates
take a look at amazon cedar for rbac implementation and see if we can overhaul our current rbac and utilize this instead. regardless, we should adjust our current rbac so that we have things lijke .view and .edit on a single perm (like can see this page > can edit it or can view it)

leahpeker:
## Overview

Investigation spike for issue #541 (`explore` label — findings only, no application code).

The issue asked two things: (1) evaluate **Amazon Cedar** as an RBAC overhaul, and (2) regardless, add **`.view`/`.edit` granularity** so a role can be granted read access to an admin surface without also granting write access.

**Findings (full spec:** `docs/superpowers/specs/2026-07-13-rbac-updates-explore.md`**):**

- **Cedar — recommend NOT adopting.** PDA's RBAC is a small, flat, static role→permission-set model (13 `PermissionKey` keys, JSON-array storage, OR-union `has_permission`). Cedar's value (ABAC / per-resource / policy-as-code) is overkill here, has no first-party Python SDK, and its supported path (AWS Verified Permissions) pulls in an AWS dependency that conflicts with PDA's Railway deployment. It would also make the frontend permission mirror *worse*. Revisit only if a genuine ABAC requirement appears.
- **`.view`/`.edit` granularity — recommend implementing in the existing flat model, phased.** No schema migration needed (keys are opaque strings in a JSON array). ~6 `manage_*` keys currently gate both GET (view) and writes (edit) on the same admin surface — these are the split candidates (surveys and documents first). The `edit_*` content-page keys already have open view + gated edit and need no `.view` counterpart. Uses the dotted `resource.view`/`resource.edit` convention already documented in the `/audit-permissions` skill.

Six open questions (split scope, backfill semantics, `/docs` route guard, role-editor UX, enforcement-helper refactor, Cedar live-fact check) are captured in the spec for a human decision before any implementation.

Closes #541

## Test plan

- [ ] N/A — documentation-only spike (no application code changed)

### DIFF
--- a/docs/superpowers/specs/2026-07-13-rbac-updates-explore.md
+++ b/docs/superpowers/specs/2026-07-13-rbac-updates-explore.md
@@ -1,0 +1,235 @@
+# Explore: RBAC updates (#541) — Findings
+
+**Date:** 2026-07-13
+**Issue:** https://github.com/ProteinDeficientsAnonymous/pda/issues/541
+**Branch / PR:** `auto-541-explore-rbac-cedar-granularity` (draft PR linked on push)
+
+## The ask
+
+Issue #541 has two parts:
+
+1. **Evaluate Amazon Cedar** as a possible overhaul of PDA's RBAC — "see if we can
+   overhaul our current rbac and utilize this instead."
+2. **Regardless of Cedar, add `.view`/`.edit` granularity** — split a single
+   permission into "can view this page" vs "can edit it," so a role can be granted
+   read access to an admin surface without also granting write access.
+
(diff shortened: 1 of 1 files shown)


leahpeker:
closing this findings spike — recommendations captured in epic #946 (implement .view/.edit granularity phased; do not adopt cedar). spec preserved on the branch for reference.

PR #739, opened 2026-07-13
refactor(users): drop transitional display_name column — PR1c (Issue 532)

leahpeker:
## overview

PR1c of the display-name → first/last split (Issue 532). PR1b (#728) is merged, so the deployed frontend no longer reads `display_name` from the API — this drops the transitional column and collapses name resolution to first/last only.

Depends on PR1a (#713, merged) + PR1b (#728, merged).

## what changed

- **Drop the column.** Remove `display_name` from `User` and `JoinRequest` (+ the `save()` sync override); add drop-column migrations. They depend transitively on the earlier backfill migrations (`0033`, `0066`), so first/last are populated before the column is removed.
- **Schemas.** `display_name` removed from every user/member/join-request input and output schema. Join-request + approve outputs now expose `first_name` / `full_name` (the PR1b admin screen already reads `fullName`).
- **Dedup resolvers.** Deleted the legacy `display_name` parse fallback from `_resolve_name_fields`, `create_user`, and the join-request submit path. `parse_display_name` is kept — the committed backfill migrations still call it against historical model state.
- **Read sites.** Every denormalized `user.display_name or phone_number` rewritten to `full_name` across events, notifications, surveys, polls, comments, audit, seed, and admin. Directory/search now order + filter on `first_name`/`last_name` (`full_name` is a property, not a column, so it can't be used in `.order_by`/`.filter`).
- **Types.** Regenerated the OpenAPI schema + frontend `types.gen.ts`.

## verification

- backend: ruff, ty, complexity, 500-line limit — clean. `1260 passed`; the 19 failures are the pre-existing SQLite-only set (`roles__permissions__contains` JSONField lookups + `pg_notify`/SSE), confirmed failing identically on `main` and green on the Postgres CI.
- frontend: tsc, eslint, prettier, openapi-types no-drift, `728` vitest passed.

## flag for review

The backend "nameless member shows the literal string `member`" fallback in the member directory and the co-host/invite search is **gone** — those surfaces now return a blank `full_name`. The member directory + member-detail frontend already fall back to the formatted phone (`fullName || formatPhone(phone)`), so a nameless member reads better than before. The one thin spot: in the co-host/invite `MemberPicker`, a nameless member who has also hidden their phone now shows a blank primary line (previously `"member"`). Rare state; calling it out rather than silently changing it.

PR #770, opened 2026-07-13
feat(public-rsvp): frontend /my-rsvps manage screen (Issue 497)

leahpeker:
## Summary

Stage 7 of the public-RSVP epic (#490). Adds a public `/my-rsvps` route (no login) that resolves a `?token=` magic link to a non-member's RSVPs and lets them manage each one.

- **`api/publicRsvp.ts`** — `useMyRsvps` (GET), `useUpdateMyRsvp` (POST), `useCancelMyRsvp` (DELETE), all passing `?token=` as a query param against `/api/community/public/my-rsvps/`.
- **`MyRsvpsScreen`** — reads the token from the URL. A 404 (unresolvable token) shows the invalid-token empty state (`this link's expired or invalid — rsvp again to get a new one`); transient failures (429/5xx/offline) show a retry message instead of misleading the holder into re-RSVPing.
- **`MyRsvpCard`** — one card per event (title, date/time, location, links), editable status reusing `RsvpStatusPicker`, a plus-one toggle when `allow_plus_ones`, and a cancel button. Each edit toasts `rsvp updated — check your email for an updated link`. Waitlisted holders re-submit as `attending` (the POST endpoint rejects `status=waitlisted`); backend messages are surfaced on 400.
- New public route wired in `router/routes.tsx`.

All UI copy is lowercase; components reused from stage 6 where practical.

## Test plan

- [x] `make agent-frontend-ci` — 736 tests pass, typecheck + lint + prettier + api-types check clean
- [x] renders the RSVP list when the token is valid
- [x] shows the invalid-token empty state when token missing / 404 (expired/revoked)
- [x] transient error shows a retry message, not the invalid-token state
- [x] editing a status posts to the manage endpoint and shows the update toast
- [x] cancel calls the delete endpoint
- [x] screen passes axe

### DIFF
--- a/frontend/src/api/publicRsvp.ts
+++ b/frontend/src/api/publicRsvp.ts
@@ -1,6 +1,8 @@
-import { useMutation } from '@tanstack/react-query';
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 
+import { mapEvent, type WireEvent } from '@/api/eventMapper';
 import type { components } from '@/api/types.gen';
+import type { Event, RsvpInputStatus } from '@/models/event';
 
 import { apiClient } from './client';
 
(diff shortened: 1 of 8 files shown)


frontend/src/api/publicRsvp.ts
```diff
@@ -23,3 +25,77 @@ export function useSubmitPublicRsvp() {
...
+
+function manageQueryKey(token: string) {
+  return ['public-my-rsvps', token] as const;
+}
+
+async function fetchMyRsvps(token: string): Promise<ManageRsvps> {
+  const { data } = await apiClient.get<WireManageOut>(MANAGE_BASE, { params: { token } });
+  const u = data.user;
+  return {
+    user: { displayName: u.display_name, email: u.email, phoneNumber: u.phone_number },
```
leahpeker: we're removing display_name (see https://github.com/ProteinDeficientsAnonymous/pda/pull/739)

this will conflict once that goes in

leahpeker: good catch — verified against #739. that PR keeps the `display_name` wire key on `PublicRsvpManageUserOut` (only swaps the value to `full_name`) and doesn't touch this frontend file, so no merge conflict or runtime break. still renamed the domain-facing field `ManageRsvps.user.displayName` → `name` (6dd4448) so the deprecated concept doesn't leak; the wire mapping still reads the `display_name` key the endpoint sends.

PR #865, opened 2026-07-15
Fix calendar list view to display event type badges

### LINKED ISSUE #830: Can’t see event type tags in list view
```
Should be able to see official tag and club tag in list view
```

## Metadata

- **Route:** `/calendar`

hopstreax:
## Summary

Fixes #830.

This change adds event type badges to calendar list/card views so users can distinguish official and club events in the calendar list view.

## Changes

- Extracted a reusable `EventBadge` component from `EventDetailScreen`
- Reused `EventBadge` in `EventCardBadges`
- Display event type badges for official and club events in list/card views
- Updated tests for the new shared component and list badge rendering

## Testing

- [x] pnpm lint
- [x] pnpm typecheck
- [x] pnpm test

Manual investigation confirmed that the backend already returns `event_type` and the frontend maps it correctly. The missing behavior was limited to the presentation layer, where the event type was used only for styling and not rendered as a textual badge.


leahpeker:
Re-using event badges is a great idea - but I'd love for the tag to be right under the event title instead of at the bottom. i know the issue wasn't clear on that

hopstreax:
Thanks! That makes sense. I agree placing the event type directly beneath the title improves the hierarchy. I'll update the layout accordingly.

hopstreax:
Thanks for the feedback! I've updated the layout so the event type badge now appears directly beneath the event title in both the agenda and day views, while keeping EventCardBadges focused on RSVP/headcount. I also pulled the badge rendering out of EventCardBadges to keep component responsibilities separated. Let me know if you'd like any other adjustments.

leahpeker:
@hopstreax FE CI failed - looks like it's just bc of imports (maybe when main was merged in)

PR #986, opened 2026-07-19
fix(rsvp): let recognized non-members rsvp to a second event inline (Issue 981)

### LINKED ISSUE #981: non-member can't rsvp for 2nd event
```
as a non-member i rsvp'd to an event, i'm able to see my rsvps and see other official events but when trying to rsvping to another official event my rsvp wont go through
```

## Metadata

- **Route:** `/events/b1686468-3f2a-4bd1-b955-ff02707afd7a`
- **User Agent:** `Mozilla/5.0 (iPhone; CPU iPhone OS 18_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.5 Mobile/15E148 Safari/604.1`

leahpeker:
## summary

A returning non-member who had already rsvp'd to one event was hitting a dead end when trying to rsvp to a later event.

**root cause:** on every event after their first rsvp, a recognized phone number routed into the phone-check "recognized" branch, which emailed the caller a manage link and showed a dead-end "check your email" message instead of the rsvp form. So the rsvp to a second event never went through.

**fix:** the rsvp-phone-check endpoint now issues a token for a recognized number and returns it. The public rsvp form unlocks inline rsvp with that token, exactly like the already-rsvpd path. Removed the now-unused email-link helper and the dead-end UI state. Updated backend + frontend tests.

Closes #981

### DIFF
--- a/backend/community/_public_rsvp_submit.py
+++ b/backend/community/_public_rsvp_submit.py
@@ -3,11 +3,10 @@ from enum import StrEnum
 
 from config.audit import audit_log
 from config.ratelimit import client_ip, rate_limit
-from django.conf import settings
 from django.db import IntegrityError, transaction
 from ninja import Router
 from ninja.responses import Status
-from notifications._email_helpers import send_rsvp_confirmation_email, send_rsvp_manage_link_email
+from notifications._email_helpers import send_rsvp_confirmation_email
 from notifications.email_sender import get_email_sender
 from pydantic import BaseModel, EmailStr, Field
 from users.models import NonMemberRsvpToken, User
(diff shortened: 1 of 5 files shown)


leahpeker:
Superseded — the phone-check flow this PR builds on (`ALREADY_RSVPD`/`RECOGNIZED` statuses, inline token-unlock for recognized non-members) was replaced by the later security-hardening series (#1000 token-leak, #1001 enumeration-oracle, #988 member-only +1s). Main now always emails non-members a manage link rather than unlocking inline RSVP via phone recognition, so this PR's approach directly conflicts with that intentional design. Closing as obsolete.

PR #988, opened 2026-07-20
Fix #987: Restrict public RSVP +1s to authenticated members

### LINKED ISSUE #987: Restrict event +1s to members only
## Context

Non-member RSVP is a fully built flow (`backend/community/_public_rsvp_submit.py`, `_public_rsvp_shared.py`, `NonMemberRsvpToken`) that runs alongside the authenticated member RSVP flow (`backend/community/_event_rsvps.py`). Both funnel through `_apply_rsvp_in_transaction`, and both currently support `has_plus_one` when the event allows it (`Event.allow_plus_ones`). Gating today is per-event only (`allow_plus_ones`, `visibility`), not per-user-type, so a non-member submitting a public RSVP can add a +1 the same as a member.

We want to restrict +1s to members only, regardless of whether the event otherwise allows public/non-member RSVPs.

## What needs to happen

Enforce that `has_plus_one` is only honored for authenticated member RSVPs. The natural seam is where `has_plus_one` is currently validated/resolved — `_resolve_rsvp_status` / `_validate_rsvp_access` in `backend/community/_event_rsvps.py` — mirrored in the non-member path (`_public_rsvp_submit.py` / `_public_rsvp_shared.py`), so a non-member's `has_plus_one` is stripped/ignored server-side (never trust the client) before `_apply_rsvp_in_transaction` runs. Update the public RSVP frontend UI to not present a +1 option to non-member submitters.

## Acceptance criteria

- [ ] A non-member (public RSVP flow, `NonMemberRsvpToken`) submitting `has_plus_one: true` has it stripped/ignored server-side; the resulting RSVP has no +1
- [ ] Member RSVP flow is unaffected — members can still add a +1 when `Event.allow_plus_ones` is true
- [ ] Public RSVP frontend no longer shows a +1 checkbox/option to non-member submitters
- [ ] Existing capacity/waitlist logic in `_apply_rsvp_in_transaction` continues to work correctly now that non-members never contribute a +1

hopstreax:
## Summary

This PR restricts **+1 RSVPs to authenticated members only**.

Previously, public (non-member) RSVP submissions could include `has_plus_one=true` whenever an event allowed +1s. This change enforces the new business rule by ensuring public RSVP requests always ignore client-supplied `has_plus_one` values.

### Backend

* Ignore `has_plus_one` in the public RSVP submission flow.
* Ensure honeypot responses also return `has_plus_one=false` for consistency.
* Added a regression test verifying that public RSVP requests cannot persist `has_plus_one=true`.

### Frontend

* Removed the +1 toggle from the public RSVP form.
* Removed the related helper text.
* Removed the unused `hasPlusOne` state.
* Public RSVP submissions now always send `has_plus_one: false`.

### Backend

* Added a regression test for public RSVP +1 enforcement.
* Ran:

```text
python -m pytest backend/tests/test_public_rsvp.py
```

Result:

* ✅ 38 tests passed

### Frontend

Ran:

```text
pnpm vitest run src/screens/events/PublicRsvpForm.test.tsx
pnpm typecheck
pnpm lint
```

All checks passed.

## Result

* ✅ Public RSVP users can no longer bring a +1.
* ✅ Crafted requests with `has_plus_one=true` are ignored by the backend.
* ✅ Member RSVP behavior remains unchanged.
* ✅ Capacity and waitlist logic continue to work through the existing shared RSVP flow.


leahpeker:
blocking chore: there's a review.diff file on this PR that should be removed

leahpeker:
oh this is my bad for not noticing before merging but please use conventional commits for PR names (the squashed commit names default to PR names):
fix(specifier): description
auto-versioning relies on conventional commit naming 

if you use the /open-pr skill it will handle this for you

Issue #999, opened 2026-07-20
epic: non-member rsvp — e2e audit bugs


leahpeker:
## non-member rsvp — end-to-end audit

epic tracking bugs found in an end-to-end audit of the non-member public rsvp feature (non-members rsvping to official public events, managing rsvps via `/my-rsvps`, commenting, and re-rsvping to more events).

**method:** 5 parallel audit passes across distinct risk dimensions (security/privacy, capacity/waitlist/races, identity resolution + email, frontend state/ux, eligibility lifecycle), each finding then adversarially re-verified against the code by an independent pass. 16 candidate findings → **12 confirmed** (2 duplicate pairs merged), 2 refuted.

### confirmed bugs

**🔴 critical**
- [x] #1000 — phone-check endpoint hands a full manage token to anyone who guesses a non-member's phone number (account takeover) — PR #1014

**🟠 high**
- [ ] #1001 — phone-check returns four distinct statuses, forming a membership/RSVP enumeration oracle
- [x] #1002 — archived member with the submitted phone bypasses the member 409 gate and 500s on token issuance — PR #1015
- [x] #1003 — _backfill_email does an unguarded save() that raises IntegrityError (500) on a unique-email collision with an archived holder — PR #1015
- [x] #1004 — waitlist promotion emails member-only event details (location + links) to a non-member after the event drops out of public-RSVP eligibility — PR #1013

**🟡 medium**
- [ ] #1005 — non-member RSVP cancellation promotes waitlisted users but never emails promoted non-members
- [ ] #1006 — final RSVP step does not client-validate name characters or phone validity, so valid-looking input 400s with a generic error

**🟢 low**
- [ ] #1007 — RECOGNIZED path lets an unauthenticated caller trigger manage-link emails to a victim by phone number alone (email bombing / unsolicited link)
- [ ] #1008 — non-member RSVP update/cancel does not broadcast capacity change, so live viewers see stale spots-left
- [ ] #1009 — phone-check reports RECOGNIZED and emails a manage link when the user's only prior RSVP is on an ineligible event, landing them on an empty /my-rsvps
- [ ] #1010 — public RSVP submit form hides all backend 400 messages behind a generic error
- [ ] #1011 — delete_my_rsvp does not gate on eligibility/cancelled/past like update_my_rsvp and member-delete, so it can promote a waitlisted non-member on a cancelled or ineligible event

### follow-up
- [ ] harden the non-member rsvp playwright e2e suite — #1012

[...]

PR #1013, opened 2026-07-21
fix(rsvp): stop leaking member-only fields in waitlist-promotion emails (Issue 1004)

### LINKED ISSUE #1004: waitlist promotion emails member-only event details (location + links) to a non-member after the event drops out of public-RSVP eligibility
part of #999

**severity:** high · **category:** security · **confidence:** high
**location:** `backend/community/_public_rsvp_shared.py:50 (_email_details) and :75 (_email_promoted_non_members)`

### what's wrong
_apply_rsvp_in_transaction / _delete_rsvp_in_transaction call promote_from_waitlist, which promotes the waitlisted non-member and returns their id; upsert_rsvp (_event_rsvps.py:214) and delete_rsvp (:234) then call _email_promoted_non_members, which builds RsvpEmailDetails via _email_details with event.location and _event_links(event) (whatsapp/partiful/other) unconditionally. The non-member receives an email containing the member-gated location and links even though the event is no longer eligible and the same fields are hidden from them in _event_out / get_event / my-rsvps. Expected: promotion email for a now-ineligible event should either be suppressed or must not include member-only location/links.

### trigger
An OFFICIAL/PUBLIC/rsvp_enabled event with max_attendees and at least one WAITLISTED non-member is edited so it is no longer public_rsvp_eligible (visibility -> MEMBERS_ONLY/INVITE_ONLY, or event_type -> COMMUNITY) while rsvp_enabled and max_attendees stay set. A member or host then changes/removes their attending RSVP (upsert_rsvp / delete_rsvp), freeing a spot.

### reproduction
1. Create an OFFICIAL, PUBLIC, ACTIVE, rsvp_enabled event with max_attendees=1 and a future start.
2. As a non-member, publicly RSVP via /public/events/{id}/rsvp/ with status=attending (fills the single spot -> ATTENDING).
3. As a second non-member, publicly RSVP status=attending — now at capacity, so _resolve_rsvp_status auto-waitlists this non-member (WAITLISTED EventRSVP row). Give this non-member a real email.
4. As a host/admin, PATCH /events/{id}/ setting event_type=COMMUNITY (leave visibility=PUBLIC, rsvp_enabled=true, max_attendees=1). Validation passes; the event is now is_public_rsvp_eligible=False but the waitlisted RSVP and max_attendees remain.
5. As the first (attending) member/non-member/host, DELETE /events/{id}/rsvp/ (or upsert to can't-go). This frees the spot -> spot_freed -> promote_from_waitlist promotes the waitlisted non-member and returns their id.
6. _email_promoted_non_members sends send_rsvp_waitlist_promoted_email with _email_details containing event.location and _event_links(event). The waitlisted non-member receives an email exposing the member-only location and whatsapp/partiful/other links for an event that is no longer public-RSVP-eligible and whose location/links are otherwise hidden from them in the app.

### evidence
_email_details always sets event_location=event.location and event_links=_event_links(event) with no eligibility guard; _email_promoted_non_members loops non-members and calls send_rsvp_waitlist_promoted_email(details=_email_details(event, user, token.token)); the promotion email template (_email_helpers.py:65 send_rsvp_waitlist_promoted_email) renders event_location/event_links. promote_from_waitlist runs whenever event.max_attendees is not None regardless of visibility/event_type.

<details><summary>verifier reasoning (adversarial re-check)</summary>

Verified the full end-to-end chain in current code. (1) A non-member CAN be waitlisted: submit_public_rsvp (_public_rsvp_submit.py:256) calls _apply_rsvp_in_transaction, and _resolve_rsvp_status (_event_rsvps.py:76-77) auto-waitlists a new attending RSVP at capacity, creating a WAITLISTED EventRSVP row for a non-member User. (2) The event can become ineligible while keeping max_attendees, rsvp_enabled, and the waitlist intact: changing event_type OFFICIAL->COMMUNITY passes validation — _is_invalid_typed_visibility (_events.py:71-73) only forces PUBLIC visibility for _PUBLIC_ONLY_TYPES={OFFICIAL, CLUB}, so COMMUNITY has no such constraint; and _apply_field_updates (_events.py:373) merely setattrs the changed fields and saves, never clearing RSVPs. After this edit is_public_rsvp_eligible (event.py:130-138) is False. (3) A freed spot triggers promotion regardless of eligibility: promote_from_waitlist (_event_helpers.py:148) guards only on `event.max_attendees is None`, with no eligibility/visibility/event_type check, so it promotes the waitlisted non-member and returns their id from upsert_rsvp/delete_rsvp. (4) The leak: upsert_rsvp (_event_rsvps.py:214) and delete_rsvp (:234) call _email_promoted_non_members, which builds RsvpEmailDetails via _email_details (_public_rsvp_shared.py:50-60) with event_location=event.location and event_links=_event_links(event) UNCONDITIONALLY. The promotion templates render both (rsvp_waitlist_promoted.txt: `where: {{ event_location|lower }}` and the event_links loop; rsvp_waitlist_promoted.html:8-9 the same). So the non-member receives member-gated location and links by email for an event that is no longer public-RSVP-eligible — the same fields _event_out/get_event/my-rsvps now hide from them. Note: two of the report's three listed triggers (visibility->MEMBERS_ONLY/INVITE_ONLY while staying OFFICIAL) are actually blocked by _is_invalid_typed_visibility, but the event_type->COMMUNITY trigger is valid and fully reproduces the defect, so the finding stands.

</details>

<sub>found via automated e2e audit of the non-member rsvp feature; adversarially verified.</sub>

leahpeker:
part of #999
fixes #1004

## the leak

`_email_promoted_non_members` (`backend/community/_public_rsvp_shared.py`) builds the "you're off the waitlist" email via `_email_details`, which sets `event_location=event.location` and `event_links=_event_links(event)` (whatsapp/partiful/other) with no eligibility check. `promote_from_waitlist` (`_event_helpers.py`) only guards on `max_attendees is not None` — no visibility/event_type check — so it fires whenever a spot frees up, even if the event has since become ineligible for public RSVP.

Repro: a PUBLIC/OFFICIAL event with `max_attendees` gets a non-member waitlisted, then a host edits `event_type` OFFICIAL → COMMUNITY (passes validation, `is_public_rsvp_eligible` goes false, but the waitlist row and `max_attendees` stay). When a spot later frees up, the promoted non-member gets an email containing the event's member-gated location and links — the same fields `_event_out`/`get_event`/`/my-rsvps` (via `public_rsvp_eligible_q`) already hide from them.

## fix

Gate on `event.is_public_rsvp_eligible` inside `_email_promoted_non_members` and strip `event_location`/`event_links` (via `dataclasses.replace`) when the event is no longer eligible, rather than suppressing the email outright — being off the waitlist is still worth telling them, and both email templates already `{% if event_location %}` / loop over `event_links`, so an empty-fields email renders cleanly (subject/body still make sense without a location).

**Scope:** the fix lives in `_email_promoted_non_members`, not in `_email_details` itself. `_email_details` is also called from `_send_confirmation_email` and `_send_updated_email` in `_public_rsvp_submit.py` / `_public_rsvp_manage.py`, but both of those run after `_load_public_rsvp_event`, which just verified eligibility in the same request — so gating there would be redundant and those call sites are correctly left unconditional.

## testing

- Added `test_ineligible_event_omits_location_and_links_from_promotion_email` to `backend/tests/test_waitlist_promotion_emails.py`: waitlists a non-member, flips `event_type` to `COMMUNITY` (event becomes ineligible), frees the spot, and asserts the promoted non-member's email (`html`/`text`) does not contain the event's location or whatsapp link.
- `cd backend && uv run pytest tests/test_waitlist_promotion_emails.py tests/test_public_rsvp.py tests/test_public_rsvp_capacity.py -q` → 47 passed
- `uv run ruff check` / `uv run ruff format --check` on both touched files → clean

Did not run the full `make agent-ci` suite per scope of this fix (single-file backend change + tests).

### DIFF
--- a/backend/community/_public_rsvp_shared.py
+++ b/backend/community/_public_rsvp_shared.py
@@ -1,4 +1,5 @@
 import logging
+from dataclasses import replace
 
 from config.audit import audit_log
 from django.conf import settings
@@ -80,7 +81,14 @@ def _log_email_failure(request, event: Event, user: User, exc: Exception) -> Non
 
 
 def _email_promoted_non_members(request, event: Event, promoted_user_ids: list[str]) -> None:
-    """Email any promoted non-members their manage link. Best-effort per user."""
+    """Email any promoted non-members their manage link. Best-effort per user.
+
+    By the time a spot frees up, the event may have dropped out of public-RSVP
+    eligibility since the non-member first waitlisted. Strip the member-only
+    location/links in that case, matching what _event_out/get_event/my-rsvps
+    already hide from them — the promotion itself is still worth telling them
+    about.
+    """
     if not promoted_user_ids:
         return
     promoted = User.objects.filter(id__in=promoted_user_ids, is_member=False, email__isnull=False)
@@ -89,9 +97,12 @@ def _email_promoted_non_members(request, event: Event, promoted_user_ids: list[s
             continue
         try:
             token = NonMemberRsvpToken.issue_or_extend(user)
+            details = _email_details(event, user, token.token)
+            if not event.is_public_rsvp_eligible:
+                details = replace(details, event_location="", event_links=[])
             result = send_rsvp_waitlist_promoted_email(
                 sender=get_email_sender(),
-                details=_email_details(event, user, token.token),
+                details=details,
             )
             if not result.success:
                 raise RuntimeError(result.error or "send returned failure")
--- a/backend/community/_public_rsvp_shared.py
+++ b/backend/community/_public_rsvp_shared.py
@@ -80,15 +80,17 @@ def _log_email_failure(request, event: Event, user: User, exc: Exception) -> Non
     )
 
 
-def _email_promoted_non_members(request, event: Event, promoted_user_ids: list[str]) -> None:
-    """Email any promoted non-members their manage link. Best-effort per user.
+def _build_promoted_email_details(event: Event, user: User, token_str: str) -> RsvpEmailDetails:
+    """_email_details, minus member-only location/links once the event is no longer
+    public-RSVP-eligible — matches what _event_out/get_event/my-rsvps already hide."""
+    details = _email_details(event, user, token_str)
+    if not event.is_public_rsvp_eligible:
+        details = replace(details, event_location="", event_links=[])
+    return details
 
-    By the time a spot frees up, the event may have dropped out of public-RSVP
-    eligibility since the non-member first waitlisted. Strip the member-only
-    location/links in that case, matching what _event_out/get_event/my-rsvps
-    already hide from them — the promotion itself is still worth telling them
-    about.
-    """
+
+def _email_promoted_non_members(request, event: Event, promoted_user_ids: list[str]) -> None:
+    """Email any promoted non-members their manage link. Best-effort per user."""
     if not promoted_user_ids:
         return
     promoted = User.objects.filter(id__in=promoted_user_ids, is_member=False, email__isnull=False)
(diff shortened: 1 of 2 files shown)


leahpeker:
Closing in favor of #1022, which removes non-member RSVPs outright when an event loses public-RSVP eligibility — that's the actual fix per the discussion on #1004. Once non-members are removed at the point of ineligibility, this PR's promotion-email redaction never has anything to redact.

PR #1019, opened 2026-07-21
fix: 1007 - stop unverified rsvp recognized-link emails

### LINKED ISSUE #1007: RECOGNIZED path lets an unauthenticated caller trigger manage-link emails to a victim by phone number alone (email bombing / unsolicited link)
part of #999

**severity:** low · **category:** security · **confidence:** high
**location:** `backend/community/_public_rsvp_submit.py:213-215`

### what's wrong
_send_recognized_login_link mints/extends a NonMemberRsvpToken and emails the victim's manage link to their address on file, with no verification the caller controls that phone or email. An attacker who knows only a target's phone number can repeatedly send scoped-link emails to the target's inbox (bounded only by 20/h per IP), and each call extends the victim's live token lifetime. Expected: only send the manage link when the caller proves ownership (as the resend endpoint does with phone+email), not off a bare phone probe.

### trigger
Unauthenticated POST to rsvp-phone-check/ with a non-member's phone number, where that non-member has an email on file and has RSVP'd to at least one event (but not this one).

### reproduction
1. Seed: non-member User V with phone +15550001111, email victim@example.com, is_member=False, who has RSVP'd to event A (so V.event_rsvps.exists() is true). Target a different eligible public event B that V has NOT RSVP'd to. 2. As an unauthenticated attacker (no token, no account), POST /api/public/events/{B.id}/rsvp-phone-check/ with body {"phone_number": "+15550001111"}. 3. Server: user resolved by phone (line 200), not a member (line 203 passes), no RSVP on event B (line 205 passes), then line 213 true → _send_recognized_login_link runs: issue_or_extend resets V's token expires_at to now+90d (users/models.py:292) and an email with the manage URL is sent to victim@example.com. Response is 200 RECOGNIZED. 4. Repeat up to 20 times/hour per IP (more across IPs) to spam V's inbox and keep resetting V's token lifetime — with no proof the attacker controls V's phone or email. Expected: like the resend endpoint, only send/extend when phone AND email are both supplied and match, and/or return RECOGNIZED without sending on a bare phone probe.

### evidence
if user.email and user.event_rsvps.exists():
    _send_recognized_login_link(request, user)  — issues token + emails manage_url with token, keyed only on the submitted phone

<details><summary>verifier reasoning (adversarial re-check)</summary>

Confirmed in current code. check_public_rsvp_phone (backend/community/_public_rsvp_submit.py:186-216) is auth=None (line 189), rate-limited only 20/h per IP (line 191). At lines 213-215 the branch `if user.email and user.event_rsvps.exists(): _send_recognized_login_link(request, user)` fires off a bare phone lookup (user resolved at line 200 by phone_number alone, no email, no ownership proof). _send_recognized_login_link (lines 159-183) calls NonMemberRsvpToken.issue_or_extend(user) then send_rsvp_manage_link_email to user.email. Two concrete harms: (1) an unauthenticated caller who knows only a non-member's phone number (where that non-member has an email on file and has RSVP'd to at least one event other than this one) triggers a real manage-link email to the victim's inbox, up to 20/h per IP and multipliable across IPs — unsolicited email / mild email-bombing; (2) issue_or_extend (users/models.py:280-295) resets the live token's expires_at to now+90d (line 292) on every call, so an attacker can keep the victim's token alive without ever seeing it, bounded only by the 180d absolute cap (max_age_cutoff, line 290). The contrast confirms the asymmetry is unintended: the sibling resend endpoint (_public_rsvp_resend.py:87-127) deliberately requires phone AND email to match (_find_non_member, lines 43-60) and rate-limits at 3/h with a neutral response — exactly the ownership proof the phone-check RECOGNIZED path skips. No guard exists elsewhere; the branch executes purely on the phone match. The token is emailed to the victim's own inbox (not leaked to the attacker), so this is not token theft — matching the reported low severity. The report's phrase 'extends the victim's live token lifetime' is accurate but capped by the 180d absolute lifetime; that caveat does not refute the email harm.

</details>

<sub>found via automated e2e audit of the non-member rsvp feature; adversarially verified.</sub>

leahpeker:
## Summary

- Fixes an unauthenticated info/email-abuse issue: the RECOGNIZED branch of `check_public_rsvp_phone` (`backend/community/_public_rsvp_submit.py`) sent a non-member's rsvp manage-link email and extended their token's expiry off a bare phone-number match, with no proof the caller controls that phone or email
- The phone-check endpoint's request schema has no email field, so there's no way to verify ownership at this step (unlike the sibling resend endpoint, which requires phone+email to match before sending). The fix removes the email/token side effect from the RECOGNIZED branch entirely — the endpoint still returns `status: "recognized"` so the frontend UX is unchanged, it just no longer fires an email or extends a token off a bare phone probe
- `already_rsvpd` branch and the resend endpoint are untouched, per the issue's scope

## Test plan

- [x] `uv run pytest backend/tests/test_public_rsvp_phone_check.py -q` — 12 passed (updated `test_recognized_sends_login_link_email` to assert no email/no token; added `test_recognized_does_not_extend_existing_token`)
- [x] `make agent-lint` — clean
- [x] `make agent-typecheck` — clean
- [ ] no frontend changes; no OpenAPI schema change (response shape unchanged), so `frontend-types` regen not needed

Closes #1007

### DIFF
--- a/backend/community/_public_rsvp_submit.py
+++ b/backend/community/_public_rsvp_submit.py
@@ -210,7 +210,6 @@ def check_public_rsvp_phone(request, event_id, payload: PublicRsvpPhoneCheckIn):
             _send_recognized_login_link(request, user)
         return Status(200, PublicRsvpPhoneCheckOut(status=PublicRsvpPhoneStatus.ALREADY_RSVPD))
     if user.email and user.event_rsvps.exists():
-        _send_recognized_login_link(request, user)
         return Status(200, PublicRsvpPhoneCheckOut(status=PublicRsvpPhoneStatus.RECOGNIZED))
     return Status(200, PublicRsvpPhoneCheckOut(status=PublicRsvpPhoneStatus.NEW))
 
--- a/backend/tests/test_public_rsvp_phone_check.py
+++ b/backend/tests/test_public_rsvp_phone_check.py
@@ -111,18 +111,34 @@ class TestPublicRsvpPhoneCheck:
         assert body["status"] == "recognized"
         assert body["rsvp_token"] == ""
 
-    def test_recognized_sends_login_link_email(self, api_client, official_event, fake_email_sender):
+    def test_recognized_sends_no_email_and_no_token_on_bare_phone_match(
+        self, api_client, official_event, fake_email_sender
+    ):
         guest = make_non_member("+14155550199", "guest@example.com")
         other_event = make_official_event(title="Other Event", start_datetime=future_iso(days=45))
         EventRSVP.objects.create(event=other_event, user=guest, status=RSVPStatus.ATTENDING)
 
+        response = post(api_client, official_event, phone_number="+14155550199")
+
+        assert response.status_code == 200
+        assert response.json()["status"] == "recognized"
+        fake_email_sender.send.assert_not_called()
+        assert not NonMemberRsvpToken.objects.filter(user=guest).exists()
+
+    def test_recognized_does_not_extend_existing_token(
+        self, api_client, official_event, fake_email_sender
+    ):
+        guest = make_non_member("+14155550199", "guest@example.com")
+        other_event = make_official_event(title="Other Event", start_datetime=future_iso(days=45))
+        EventRSVP.objects.create(event=other_event, user=guest, status=RSVPStatus.ATTENDING)
+        existing_token = NonMemberRsvpToken.issue_or_extend(guest)
+        original_expires_at = existing_token.expires_at
+
         post(api_client, official_event, phone_number="+14155550199")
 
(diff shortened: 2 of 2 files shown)


leahpeker:
superseded by #1018 — that branch addresses the same unverified-email abuse (Issue 1007) with a 5-minute per-user cooldown on the recognized-link email (`_send_recognized_login_link`), which stops inbox-spam and token-expiry-grief off a bare phone-number probe regardless of IP rotation. #1018 keeps emailing existing non-members (email is the only channel we have for them) but gates it behind that cooldown, rather than dropping the email for the recognized-elsewhere case. closing in favor of #1018.

PR #1156, opened 2026-07-24
docs(notifications): spec per-type notification preferences (email + bell)

leahpeker:
## Summary
- adds `.claude/docs/spec-push-and-notification-prefs.md` — phased spec for push notifications (web push first, Capacitor/FCM later) and per-type notification opt-outs
- key design points: opt-out-only `NotificationOptOut` model enforced once at notification creation (SSE + push inherit prefs for free), grouped settings toggles with a separate email toggle for types that also send email (event invites, co-host invites, check-in nudges), and splitting `event_comment` into `event_comment_hosting` / `event_comment_attending` — the attending fan-out is net-new (attendees currently get no comment notification)
- builds on `spike-native-push.md` and `app-store-readiness.md`; supersedes the spike's raw-APNs sender choice with FCM

## Test plan
- [ ] docs-only change — review the spec content, especially the open questions section


leahpeker:
## rescoped: push cut, email + bell only

Confirmed there is no push infrastructure in the repo — no service worker, no Capacitor shell,
no FCM/APNs dependency. Push isn't wanted, so phases 3–5 are removed and the file is renamed
(`spec-push-and-notification-prefs.md` → `spec-notification-prefs.md`). The push research
survives in `spike-native-push.md` / `app-store-readiness.md` if it's ever revived.

Phases 1–2 stand: per-type opt-outs across the bell and email, then the hosting/attending
comment split.

**Decisions folded in:**
- `checkin_nudge` and `rsvp_declined_note` are host-facing and now **always on** — a host needs
  to know who can't make it and that check-in is open. Both dropped from the settings UI table.
- Bell and email are independently toggleable per type.
- Milestone-reminder and event-blast email keys stay deferred.
- Push-title question is moot and removed.

**One open question left**, which overlaps Issue 1143: RSVP confirmation emails are scoped to
send on all status transitions including waitlist promotion, so they need a `NotificationType`
to hang an email toggle on. `waitlist_promoted` exists (with no producer at all today), but a
plain "you're going" confirmation has no type yet. That gets decided here or in 1143.

PR #1175, opened 2026-07-31
fix(nav): link signed-out ticket button to login (Issue 1166)

### LINKED ISSUE #1166: Ticket button not oressing
## Context

Feedback submitted from `/profile` — when signed out, pressing the ticket button does nothing instead of showing the page with a login modal. More broadly, whenever the login modal is shown on a page, the footer/tab navigation should still be visible and usable.

## What needs to happen

1. Ticket button: when no auth token is present, don't silently no-op — show the underlying page with the login modal.
2. Whenever the login modal is shown on any page, the footer and tab navigation should remain visible and clickable underneath/alongside it.

## Acceptance criteria

- [ ] Signed-out ticket button press shows the page + login modal instead of doing nothing
- [ ] Login modal, wherever shown, does not hide or block the footer/tab navigation

---

## Original

```
When I'm signed out pressing the ticket button doesn't do anything 

If it can't find a token, it shouldn't block me - I should still see the page with the login modal 

And in fact whenever I see the login modal on pages I should also see a footer and be able to click to other tabs
```

## Metadata

- **Route:** `/profile`

leahpeker:
## Summary

- Signed-out users pressing the ticket / my-rsvps nav item hit a silent no-op — with no auth token and no stored RSVP token, `BottomNav` rendered a non-interactive `div` instead of a link, so nothing happened.
- `myRsvpsDestination` now returns `/login?redirect=/events/mine` for that case, so the item renders as a normal `NavLink` — the same mechanism used everywhere else in the app to send signed-out users to auth (route guards, event detail, poll card, public rsvp form).

Closes #1166

## Test plan

- [x] `tsc --noEmit` clean
- [x] eslint clean
- [x] prettier clean
- [x] full frontend suite passes
- [ ] Signed out, press the ticket icon in the bottom nav — navigates to `/login?redirect=%2Fevents%2Fmine` instead of doing nothing
- [ ] Signing in from there lands on `/events/mine`


leahpeker:
Switched approach in `4b893818` — this PR no longer adds a login modal.

The modal (`LoginPromptDialog`) was a second sign-in surface: it forked `LoginScreen`'s phone → `checkPhone` → password flow, and bailed out to `/login` anyway for non-member and pending-review cases because that copy only exists on the screen. It also needed a `Dialog` z-index/inset change so it would sit above the nav.

Everywhere else in the app, signed-out → auth is a navigation to `/login`:

| Location | Mechanism |
|---|---|
| `auth/guards.tsx:109,125,147` | `<Navigate to={/login?redirect=...} />` |
| `EventDetailScreen.tsx:196,267` | `<Link to="/login">` |
| `poll/EventPollCard.tsx:79` | `<Link to="/login">` |
| `PublicRsvpForm.tsx:166` | `navigate('/login')` |

So the ticket button now does the same: `<NavLink to="/login?redirect=%2Fevents%2Fmine">`. That also deletes the dead `NavItem` branch that rendered a non-interactive `div` — the actual cause of "pressing it does nothing".

`/login` renders outside `AppShell`, which is why it had no bottom nav and why the modal was reached for originally. `AuthLayout` now takes an opt-in `showBottomNav`, enabled on `LoginScreen` only — onboarding, new-password, and consent share that layout and are required flows where nav would let users wander out mid-step.

Net **+50 / −207**, one sign-in surface instead of two. `Dialog.tsx` is back to its original styling.

Verified: `tsc --noEmit` clean, eslint clean, prettier clean, full frontend suite 1028 passed / 1 skipped.

PR #1181, opened 2026-08-01
feat(calendar): let members filter event types in their subscription feed (Issue 1169)

### LINKED ISSUE #1169: Cal subscription checkbox
## Context

Feedback submitted from `/calendar` — the app distinguishes event types (official, club, "my events"/hosting or RSVP'd, invited), but the calendar-subscription feed doesn't let a member filter which types are included.

## What needs to happen

Add a checkbox selection (in the settings page where calendar subscription lives) letting a member choose which event types are included in their calendar subscription feed.

## Acceptance criteria

- [ ] Settings page offers per-event-type checkboxes for calendar subscription
- [ ] The subscribed calendar feed reflects the selected types only

## Open Questions

- Do selections take effect immediately on the existing subscription URL, or does changing them require re-subscribing?

---

## Original

```
We have diff event types 
Official 
Club 
My events (hosting or RSVPd to)
Invited 

Can we make a checkbox selection for which events show up in your calendar subscription (via settings page where subscribe option is)
```

## Metadata

- **Route:** `/calendar`

leahpeker:
## Summary

- Adds per-event-type checkboxes to the calendar section of settings so a member can choose which of official / club / community events appear in their subscribed `.ics` feed.
- **Open question resolved: changes take effect immediately on the existing subscription URL — no re-subscribing.** The codebase already had the seam: a `calendar_feed_scope` preference on `User`, patched via `PATCH /api/auth/me/` and honored by the feed endpoint, which resolves a `User` from the `calendar_token` in the URL. Extended that rather than adding a parallel mechanism. Encoding filters in the URL would break every existing subscription on each checkbox change.
- **Stores exclusions, not inclusions** (`calendar_feed_excluded_types`). An empty default preserves current behavior for every existing subscriber with no data migration, and any event type added later shows up in existing feeds instead of being silently dropped. The UI still presents it as inclusion (checked = included), as the issue asked.
- `MePatchIn` types the field as `list[Literal["official","community","club"]]` so unknown values are rejected at the trust boundary, with a validator that dedupes and sorts.
- Migration: `users/0043_user_calendar_feed_excluded_types.py`.

Closes #1169

## Test plan

- [ ] Backend: feed includes all types by default; a deselected type is dropped; deselecting all yields an empty feed; preference persists via `PATCH /me/`; an unknown type is rejected 422
- [ ] Frontend: all types render checked by default; unchecking sends the exclusion; re-checking removes it
- [ ] An existing subscriber's feed URL keeps working and reflects changes without re-subscribing
- [ ] Verified: the 10 test files touched by the new required `User` field all pass (119 tests) — the field ripple was applied by scripted insertion, so this was re-run explicitly


leahpeker:
Merge conflicts with `main` need manual resolution. Auto-resolution was skipped to avoid a risky merge.

10 conflicted files, all test files:

- `frontend/src/auth/store.test.ts`
- `frontend/src/screens/admin/MembersScreen.test.tsx`
- `frontend/src/screens/auth/OnboardingProfileStep.test.tsx`
- `frontend/src/screens/events/EventDetailScreen.test.tsx`
- `frontend/src/screens/events/EventMemberSection.test.tsx`
- `frontend/src/screens/public/FaqScreen.test.tsx`
- `frontend/src/screens/public/GuidelinesScreen.test.tsx`
- `frontend/src/screens/public/HomeScreen.test.tsx`
- `frontend/src/screens/public/InstallAppScreen.test.tsx`
- `frontend/src/screens/settings/SettingsScreen.test.tsx`

Cause: this branch adds `calendarFeedExcludedTypes: []` to inline user mock objects across these files, while `main` independently restructured the same mocks (several appear to have moved to the shared `frontend/src/test/fixtures.ts` helper). The conflicts are logic-file conflicts rather than mechanical import unions, and the count exceeds the safe auto-resolve threshold, so the branch was left untouched.

Likely the cleanest path is to rebase onto `main` and re-apply the field via the shared fixture in `frontend/src/test/fixtures.ts` rather than per-file mocks, which should collapse most of these conflicts.

PR #1185, opened 2026-08-01
fix: 1182 - remove co-host picker from event edit form

### LINKED ISSUE #1182: Event edit form's co-host picker is empty and silently replaces the roster
## Context

On the event edit form, the co-host picker starts empty on **every** event, even when the event has accepted co-hosts and pending co-host invites. Nothing in that section reflects the actual roster — not the creator, not accepted co-hosts, not invited-but-unaccepted hosts.

This is intentional as written. `EventForm.tsx` explains it:

> `coHosts` is picker-driven local state. We can't fully reconstruct `MemberSearchResult` from `existing` (no phone numbers in the event-out wire shape, by design), so it starts empty even on edit.

A `coHostsDirty` flag compensates: while the picker is untouched, `coHostIds` is omitted from the PATCH so the backend preserves the existing list.

The problem is what happens once you *do* touch it. `sync_cohost_invites` treats `co_host_ids` as a **replacement**, not an append — any existing invite whose user isn't in the submitted set is rescinded, and accepted co-hosts are removed from `event.co_hosts`:

```python
for user_id, invite in existing_by_user.items():
    if user_id in next_ids:
        continue
    if _rescind_invite(invite):
        accepted_co_host_ids_to_remove.append(user_id)
if accepted_co_host_ids_to_remove:
    event.co_hosts.remove(*accepted_co_host_ids_to_remove)
```

So adding one person to a picker that *looks* empty submits `[thatOnePerson]` and silently wipes every other host and pending invite. The empty picker makes this easy to trigger by accident — the UI gives no indication there was anything to lose.

Note the creator is doubly invisible here: the picker is empty, and `_next_ids_excluding_creator` filters the creator out of `co_host_ids` regardless, since the roster is managed on the event detail page.

Found while working on Issue 425. Pre-existing on `main` — not introduced by that work.

## What needs to happen

Pick one of two directions:

**A. Show the real roster.** Add a lightweight member-summary shape (id, name, photo — no phone) to the event payload so the picker can hydrate with current accepted co-hosts and pending invitees, visually distinguishing pending from accepted. Removes the footgun by making the replace semantics visible and correct.

**B. Take the picker out of edit.** Disable/remove co-host management on the edit form and point to the event detail page, which is already the canonical place to add, rescind, kick, and step down. Smallest change, and removes the data-loss path entirely.

B is the smaller fix and matches where roster management already lives; A is better if editing hosts from the form is genuinely wanted.

## Acceptance criteria

- [ ] Editing an event and saving without touching the co-host picker leaves accepted co-hosts and pending invites unchanged (already true — add a regression test)
- [ ] It is not possible to silently drop existing co-hosts or pending invites by adding one person to the picker
- [ ] If the picker stays on the edit form, it reflects the event's current accepted co-hosts and pending invited hosts on load
- [ ] Pending invited hosts are visually distinguishable from accepted co-hosts wherever they are shown

leahpeker:
## Overview

On the event edit form, the co-host picker always started empty — even when the event already had accepted co-hosts and pending invites — because the form can't reconstruct member details from the event's wire shape (no phone numbers, by design). A `coHostsDirty` flag was meant to compensate by omitting `coHostIds` from the PATCH while the picker was untouched, but as soon as a host touched the picker to add one person, the backend's `sync_cohost_invites` treated the submitted list as a **replacement** — silently rescinding every other pending invite and removing every other accepted co-host, including the creator.

This implements direction B from the issue (the smaller, recommended fix): the co-host picker is removed from the event **edit** form entirely. Co-host management (invite, rescind, kick, step down) already lives on the event detail page — that stays the single source of truth. The picker still works as before on **event creation**, where there's no existing roster to lose.

Concretely:
- `EventForm.tsx`: removed the `coHostsDirty` tracking flag; on edit, `coHostIds` is now unconditionally omitted from the PATCH body regardless of picker state. The "hosts" card renders the picker only when creating a new event; on edit it shows read-only text pointing to the event page.
- Backend required no changes — `_apply_field_updates` already treats a missing `co_host_ids` key as "leave existing invites/co-hosts alone," it only acts when the key is present.
- Added `backend/tests/test_event_edit_preserves_cohosts.py`: regression coverage proving a PATCH without `co_host_ids` preserves both accepted co-hosts and pending invites (and that an explicit empty list still works for intentional removal).

## Test plan

- [x] `make agent-frontend-typecheck` — clean
- [x] `make agent-frontend-lint` — clean
- [x] New regression tests (`test_event_edit_preserves_cohosts.py`) — 3/3 pass
- [x] Existing cohost-related backend tests (invites, remove, past-guard, invite email) — 58/58 pass
- [x] Existing EventForm-adjacent frontend tests — 35/35 pass
- [ ] Manual: edit an event with existing co-hosts/pending invites and confirm saving without touching hosts leaves the roster untouched
- [ ] Manual: confirm the edit form's hosts card now reads "manage co-hosts on the event page" instead of showing an empty picker

Closes #1182

### DIFF
--- a/backend/tests/test_event_edit_preserves_cohosts.py
+++ b/backend/tests/test_event_edit_preserves_cohosts.py
@@ -35,9 +35,7 @@ def cohost(db) -> User:
 
 @pytest.mark.django_db
 class TestEventEditPreservesCohosts:
-    def test_patch_without_cohost_ids_preserves_accepted_cohosts(
-        self, api_client, creator, cohost
-    ):
+    def test_patch_without_cohost_ids_preserves_accepted_cohosts(self, api_client, creator, cohost):
         event_id = Event.objects.create(
             title="Potluck",
             start_datetime=future_iso(days=30),
@@ -68,9 +66,7 @@ class TestEventEditPreservesCohosts:
         invite.refresh_from_db()
         assert invite.status == CoHostInviteStatus.ACCEPTED
 
-    def test_patch_without_cohost_ids_preserves_pending_invites(
-        self, api_client, creator, cohost
-    ):
+    def test_patch_without_cohost_ids_preserves_pending_invites(self, api_client, creator, cohost):
         event_id = Event.objects.create(
             title="Potluck",
             start_datetime=future_iso(days=30),
--- a/backend/tests/test_event_edit_preserves_cohosts.py
+++ b/backend/tests/test_event_edit_preserves_cohosts.py
@@ -1,0 +1,132 @@
+import json
+
+import pytest
+from community.models import Event, EventCoHostInvite, EventStatus
+from community.models.choices import CoHostInviteStatus
+from ninja_jwt.tokens import RefreshToken
+from users.models import User
+
+from tests.conftest import future_iso
+
+
+def _make_user(phone: str, name: str = "Member", email: str | None = "") -> User:
+    return User.objects.create_user(
+        phone_number=phone,
+        password="testpass123",
+        first_name=name,
+        email=email,
+    )
+
+
+def _auth_headers(user: User) -> dict:
+    refresh = RefreshToken.for_user(user)
+    return {"HTTP_AUTHORIZATION": f"Bearer {refresh.access_token}"}  # type: ignore
+
+
+@pytest.fixture
+def creator(db) -> User:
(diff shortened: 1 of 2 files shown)


leahpeker:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #1191, opened 2026-08-01
feat: 1183 - dev-only tab for quickly creating test events

### LINKED ISSUE #1183: Add dev-only tab for quickly creating test events
## Context

Manually creating events for testing is tedious — seeding realistic events on the local dev server and on staging currently means either hand-filling the event form repeatedly or re-running a seed command. A lightweight in-app "dev" tab would make it fast to spin up test events while working on calendar/event/RSVP features.

This is a dev-tooling affordance only: it must appear on local dev and staging, and must never be reachable in production.

## What needs to happen

- Add a dev-only tab/screen in the frontend that exposes a quick "create test event" action.
- Gate visibility on environment so it renders only on dev and staging.
- Backing endpoint (or reuse of the existing event-create endpoint) must be equally gated on the backend — hiding the tab in the UI is not the gate.
- Keep it minimal: sensible defaults (title, near-future start/end, location) with a small number of overridable fields, plus a quick way to generate several events at once if that's cheap.

## Acceptance criteria

- [ ] a dev tab is visible in the app on local dev and on staging
- [ ] the tab is not visible, and its backing endpoint is not usable, in production
- [ ] creating a test event from the tab produces a real event that shows up on the calendar
- [ ] created events are obviously identifiable as test data (e.g. a recognizable title prefix)
- [ ] all user-facing text in the new UI is lowercase, per project convention

leahpeker:
## Overview

Implements Issue 1183: a dev-only tab for quickly creating fully-populated test events — always accessible, gated by environment and permission.

**Backend** (`backend/community/_dev_tools.py` + `_dev_tools_populate.py` + `_dev_tools_content.py`): `POST /api/community/dev/test-events/`, gated by `is_seed_allowed` (local + staging only, 404 elsewhere) **and** the `manage_events` permission (404, not 403 — doesn't confirm the route exists to unauthorized members).

Configurable per event:
- **Status/type**: past, canceled, official, club (mutually exclusive with official), make-me-a-host (default off — the creator is not auto-added as co-host unless requested)
- **Attendees**: going/maybe/can't-go/invited counts (default 5 each), accepted/invited co-host counts. Fillers are drawn from the existing user pool first, created (realistic random names, reserved `+1555` phone range) only when the pool runs out
- **Non-member RSVPs**: a separate `non_member_going_count`, only reachable when the event is official — mirrors `Event.is_public_rsvp_eligible`'s real eligibility gate. Fixed a bug during review where this was originally an all-or-nothing toggle that replaced every filler with a non-member instead of adding alongside members
- **Capacity**: `max_attendees` + `allow_plus_ones`; when `going` exceeds `max_attendees`, overflow lands as `WAITLISTED` instead of `ATTENDING`, mirroring the real capacity-check outcome
- **Cost fields, visibility, rsvp_enabled** (default on)
- A placeholder photo (solid-color JPEG via Pillow) so events don't render blank

**Frontend**: floating corner button in `AppShell`, self-gates on auth + non-production environment. Opens a scrollable overrides panel; on create, redirects straight to the new event instead of just toasting.

## Fixed during review

- Dialog layout: the overrides panel had a hardcoded `max-h-80` instead of flexing to the dialog's available height, which visually cut off the last few fields (including focus rings) above the create button.
- Non-member fillers were all-or-nothing (see above).
- The endpoint had no permission check beyond auth — now requires `manage_events`.
- The creator was always auto-added as a co-host via the `seed_creator_as_host` signal; now opt-in via a toggle.

## Testing

- Backend: `backend/tests/test_dev_tools.py` (20 tests) — permission gate, env gate, every toggle/field, waitlist overflow, non-member add-alongside-members, host toggle, filler pool reuse.
- Frontend: `DevTestEventsButton.test.tsx` (5 tests) — env/auth gating, default payload, non-member field disabled state.
- Manually clicked through against a running local dev server: created single and bulk-attendee events, confirmed draft/active/canceled/official/club paths, verified the non-member-fillers bug and its fix, verified the co-host default-off fix, confirmed dialog layout renders cleanly with no cutoff.
- `make agent-lint`, `agent-typecheck`, `agent-complexity`, full backend + frontend test suites all pass (backend: 1835 passed, only the 6 pre-existing SQLite-only failures unrelated to this change — `test_sse.py`/`test_rsvp_capacity_race.py`, both need real Postgres).
- Branch kept current with `origin/main` (merged twice during review) to pick up the event-slug migration and a manage_events permission fix.

## Test plan

- [x] Backend tests for every configurable field, permission gate, env gate
- [x] Frontend tests for gating and default payload
- [x] Manual click-through on local dev, including regression-testing the two bugs found and fixed above
- [x] Full `make agent-ci`-equivalent pre-PR checks

Closes #1183


leahpeker:
Design decisions resolved with the issue author — spec updated (`55248243`). The Open Questions section is now a **Decisions** section; one new question remains.

| # | Question | Decision |
|---|---|---|
| 1 | Surface | **Floating corner button**, not a nav tab |
| 2 | Permission | **Auth only** (`gated_jwt`), no `PermissionKey` |
| 3 | Prod response | **404** |
| 4 | OpenAPI disclosure | **Accepted** — keeps CI's schema check green |
| 5 | Bulk create | **In scope**, unique titles required |
| 6 | Cleanup | **In scope**, delete-by-title-prefix |
| 7 | Default status | **ACTIVE**, overrides behind a disclosure |

**On #1** — the precedent turned out cleaner than the original recommendation. `AppShell.tsx:53` already renders `<FeedbackButton />` as a floating `position: fixed` element outside the nav grid, and that component **self-gates internally** (`FeedbackButton.tsx:17-19`) — it computes its own visibility and returns null, so `AppShell` renders it unconditionally. A dev button copies that shape exactly: no route guard, no `Tile.perm` widening (`AdminHubScreen.tsx:11`), and no `grid-cols-5` reflow (`BottomNav.tsx:26`).

**On #5** — worth calling out for whoever implements: `seed_events()` uses `get_or_create(title=...)` (`_seed_shared.py:33`), so N events with identical titles silently collapse into **one**. Bulk create needs the `secrets.token_hex(4)` suffix idiom from `e2e_seed.py:29`.

**On #3** — 404 departs from the house `raise_validation(..., status_code=403)` pattern (`_feature_flags.py:30-46`), so it's worth a one-line comment at the call site to stop a future reader "fixing" it back to 403.

**New open question:** default `visibility` for test events. ACTIVE + `PUBLIC` (the `SeedEvent` default, `_seed_shared.py:19`) puts test events on staging's **public** calendar, visible to anonymous visitors. `MEMBERS_ONLY` still satisfies the "shows up on the calendar" criterion for a logged-in dev. Leaning `MEMBERS_ONLY` as default with `PUBLIC` as an override — flagged, not blocking.

PR #1213, opened 2026-08-02
feat: 1045 - payment confirmation gate (backend)

leahpeker:
## Overview

Backend half of the payment-confirmation gate for Issue 1045 — users must confirm they paid before an RSVP to a paid event is finalized.

**Ships dark.** Every gate here is behind the existing `EVENT_PAYMENT_CONFIRMATION` feature flag (added in #1132, defaults `False`), so merging this changes nothing in production until the flag is toggled in the admin UI. `tests/test_payment_confirmation_flag_off.py` — the flag-off baseline from #1132 — still passes unmodified, which is the proof.

### What's here

- **`EventRSVP.paid_confirmed_at`** — nullable timestamp mirroring the existing `cancelled_at` pattern. Migration `0088`, no backfill.
- **`community/_rsvp_payment.py`** — two helpers. `event_requires_payment_confirmation()` (price set AND at least one of venmo/cashapp/zelle) and `requires_payment_gate()` (the enforcement question, flag-aware).
- **Enforcement in `_apply_rsvp_in_transaction`** — the single shared write path all three RSVP flows funnel through, so member, new-person, and token-holder paths are covered by one guard.
- **`paid_confirmed` on three input schemas** — `RSVPIn`, `PublicRsvpIn`, `PublicRsvpManageIn`.
- **Payment links now visible to anonymous viewers on public-RSVP-eligible events** — otherwise a stranger hits a "did you pay?" prompt with no way to pay. Scoped to `Event.is_public_rsvp_eligible` (already requires official + active + public + rsvp_enabled + not past); community/club/draft/past events keep them member-gated.

### Behavior

| Rule | Behavior |
|---|---|
| Which statuses gate | Only `attending` |
| When it fires | Transition *into* attending only — an already-attending member toggling a +1 or saving a comment is never re-prompted |
| "Has a cost" | `price` non-empty AND at least one payment method set |
| Existing RSVPs | No backfill; the transition rule protects them |
| Host/admin on behalf of a guest | Exempt — `_apply_host_rsvp_in_transaction` is a separate path, untouched |
| Confirmation persistence | Once stamped, preserved across later status changes |

### Notes for review

- The guard sits *before* the no-op early-return in `_apply_rsvp_in_transaction`, so an identical re-submit still short-circuits as it did before.
- `event_requires_payment_confirmation` deliberately does **not** consult the flag — the serializer needs "does this event cost money" regardless of enforcement.

Frontend (`PaymentConfirmStep`, the three gate wirings, error copy) follows in a second PR.

Part of Issue 1045.

## Test plan

- [x] `tests/test_rsvp_payment_gate.py` — 32 new tests across the helpers, all three endpoints, and serializer visibility
- [x] `tests/test_payment_confirmation_flag_off.py` — the #1132 flag-off baseline passes unmodified
- [x] Mutation-checked: disabling the guard fails all 5 enforcement tests, confirming they have teeth
- [x] RSVP regressions green — `test_rsvp.py`, `test_rsvp_comment.py`, `test_host_rsvp.py`, `test_event_capacity.py`, `test_public_rsvp*.py`, `test_public_rsvp_token_exposure.py`
- [x] Full backend suite: 1869 passed. The 6 failures (`test_sse.py`, `test_rsvp_capacity_race.py`) are pre-existing SQLite/`pg_notify` limitations — verified identical on a clean `origin/main` worktree
- [x] `make agent-lint`, `make agent-typecheck`, `make agent-complexity` clean


leahpeker:
Split into a 5-PR stack for reviewability. Every payment-related source file across the stack is **byte-identical** to this branch — verified with `git diff` per file — so no behavior was changed in the split, only the review boundaries.

| PR | Scope | Total | Hand-written |
|---|---|---|---|
| #1224 | model fields + helper predicates (no-op at runtime) | 321 | 318 |
| #1225 | the gate itself, on the shared rsvp write path | 549 | 519 |
| #1226 | serializer visibility (2 commits: pure extraction, then the change) | 301 | 271 |
| #1227 | payment-aware waitlist promotion notices | 180 | 180 |
| #1228 | host confirm/revoke endpoint + notification | 657 | 466 |

"Hand-written" excludes `openapi_schema.json` and `types.gen.ts`, regenerated per-PR rather than carried as one blob.

Two things this surfaced that were buried in the 1,873-line diff:

- **#1228 is a separate feature** — a new endpoint, a new notification type, and a migration that this PR's description didn't mention at all.
- **#1226's `_event_list_out` extraction** was 96 lines of diff noise obscuring a ~6-line behavior change; it's now its own commit.

Suggest closing this one in favor of the stack.

leahpeker:
Implemented by PR #1239 - payment confirmation gate frontend

PR #1234, opened 2026-08-03
feat: add an opt-in cached_flags() scope for feature flags

leahpeker:
## Overview

Groundwork for the Issue 1045 payment stack: `resolve_flags()` runs a query on **every** call, so anything checking a flag per item in a loop pays a query per item.

`cached_flags()` is a context manager that serves one query for the duration of a bounded block.

```python
with cached_flags():
    return [serialize(e) for e in events]   # one query, not one per event
```

### Why opt-in rather than ambient

A feature flag is a kill switch. A cache that applied everywhere would let a management command or worker loop hold a stale value for the life of the process — `send_attendance_reminders` and `send_checkin_nudges` both check a flag per run. Outside the scope, reads stay uncached and a toggle is visible immediately.

I tried the ambient version first (cache keyed to `request_started`/`request_finished` signals) and it broke both command test suites exactly this way — the cache leaked into contexts that never fire request signals. That failure is what the opt-in design is responding to.

### Notes for review

- `Local()` rather than a module global, so threads and async tasks can't share an entry.
- Nesting is safe — the outermost block owns the entry.
- The scope deliberately **freezes** flags for its lifetime; that's the tradeoff for one query, and it's tested explicitly.
- `clear_flag_cache()` is exported for a writer that needs to re-read inside a scope.
- No call sites adopt it here. The payment stack (#1224 → #1232) is the first consumer.

## Test plan

- [x] `tests/test_feature_flag_cache.py` — 11 tests: one-query-inside, uncached-outside, scope exit, nesting, exception unwinding, mid-scope freeze, explicit clear
- [x] Mutation-checked against the earlier ambient design
- [x] Full backend suite: **1863 passed**. The 6 failures (`test_sse.py` ×5, `test_rsvp_capacity_race.py`) are the pre-existing SQLite/`pg_notify` limitations
- [x] `make agent-lint`, `agent-typecheck`, `agent-complexity` clean

Part of Issue 1045.

### DIFF
--- a/backend/community/models/__init__.py
+++ b/backend/community/models/__init__.py
@@ -39,7 +39,13 @@ from community.models.event import (
     event_lookup_q,
     parse_event_ref,
 )
-from community.models.feature_flag import FeatureFlagState, flag_enabled, resolve_flags
+from community.models.feature_flag import (
+    FeatureFlagState,
+    cached_flags,
+    clear_flag_cache,
+    flag_enabled,
+    resolve_flags,
+)
 from community.models.join_form import JoinFormQuestion, JoinRequest
 from community.models.poll import EventPoll, PollOption, PollVote
 from community.models.survey import (
@@ -89,6 +95,8 @@ __all__ = [
     "event_lookup_q",
     "parse_event_ref",
     "FeatureFlagState",
+    "cached_flags",
+    "clear_flag_cache",
     "flag_enabled",
     "resolve_flags",
     "JoinFormQuestion",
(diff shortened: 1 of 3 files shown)


leahpeker:
Closing as unadopted: cached_flags() was built ahead of a per-item flag-check call site, but tracing every current flag_enabled()/resolve_flags() caller shows none run in a loop — event serialization checks once per event, and the report/reminder commands gate once before their loop starts. The stated first consumer (#1224 → #1232) merged without a per-item flag read. No live N+1 to justify the cache today; revisit if/when a genuine list-serialization flag check shows up.

PR #1239, opened 2026-08-03
feat: 1045 - payment confirmation gate (frontend)

leahpeker:
## Overview

Frontend half of the payment-confirmation gate for Issue 1045 — the last piece of the backend stack (#1213, #1227, #1228, #1232) merged to `main`, so this branches directly off it.

**Ships dark alongside the backend.** `EVENT_PAYMENT_CONFIRMATION` defaults `False`; every gate here checks it via `useFlag(Feature.EventPaymentConfirmation)` before showing the payment step, so merging changes nothing until the flag is toggled.

### What's here

- **Model + mapper** — `Event.myPaidConfirmed`, `EventGuest.paidConfirmed` mapped from the backend's `my_paid_confirmed`/`paid_confirmed`. The hand-maintained `WireEvent`/`WireGuest` types in `eventMapper.ts` needed the fields added by hand (they aren't OpenAPI-generated).
- **`eventCost.ts`** — `eventRequiresPaymentConfirmation()` (price + at least one payment method), and `formatPrice()` moved here from `EventMemberSection.tsx` so both the cost card and the new payment step share it.
- **`usePaymentGate(event)`** — one hook, reused by all three surfaces: flag on AND status is attending AND event costs money AND the viewer hasn't already confirmed.
- **`PaymentConfirmStep`** — shared UI: price, normalized venmo/cashapp links (via the existing `toVenmoUrl`/`toCashAppUrl` handle normalizers — a bare `@handle` needs converting, plain `ensureHttps` isn't enough), zelle as text, confirm/back.
- **Three gate wirings**, matching the four-surface bypass map from the original exploration:
  - `RsvpBox` — member + token-holder create/edit dialog
  - `PublicRsvpForm` — brand-new person's step machine
  - `PublicRsvpCard` — token holder's quick-edit, which fires its mutation immediately with **no dialog at all** today. This was the one the original issue text didn't mention and would've stayed a silent bypass.
- **Error copy** for `event.payment_confirmation_required` on all three surfaces, for the stale-client case.

### Notes for review

- `RsvpBox` and `PublicRsvpCard` both had a hardcoded `paid_confirmed: false` / `paid_confirmed: false` already in their mutation call — leftover stubs from earlier work on this stack that would have silently defeated confirmation on those paths. Fixed as part of threading the real value through.
- Rebuilt this branch's history once: it was originally stacked on the backend branch pre-merge, and rebasing onto `origin/main` post-squash-merge produced bogus conflicts in backend files I never touched (the same content reapplying against itself). Reset to `origin/main` and cherry-picked the 10 real frontend commits instead — diff is frontend-only, verified via `git diff origin/main --stat`.
- Running the full test suite (not just changed files) surfaced 4 test files that render `PublicRsvpCard`/`PublicRsvpForm` without a `QueryClientProvider` and without mocking `@/api/featureFlags` — `usePaymentGate`'s `useFlag` call threw `No QueryClient set`. Fixed by mocking the flag off in those files, consistent with how every other test in this PR mocks it.

Part of Issue 1045.

## Test plan

- [x] New coverage: `eventMapper`, `eventCost`, `usePaymentGate`, `PaymentConfirmStep`, plus gate-specific describe blocks in `RsvpBox`, `PublicRsvpForm`, `PublicRsvpCard` — free event / paid event / flag-off / already-confirmed / non-attending-status all covered per surface
- [x] Full frontend suite: **138 files, 1128 tests passing**, 1 pre-existing skip
- [x] `pnpm tsc -b --noEmit` clean (project-references-aware; the bare `tsc --noEmit` silently skips `src/test/**` and `*.test.tsx` and would have missed several of the wire-type fixes below)
- [x] `make agent-frontend-lint`, `make agent-frontend-format-check` clean
- [x] Verified `git diff origin/main --stat` touches only `frontend/` — no backend files carried over from the rebuild

### DIFF
--- a/frontend/src/api/eventMapper.test.ts
+++ b/frontend/src/api/eventMapper.test.ts
@@ -37,8 +37,8 @@ describe('mapEvent', () => {
         ],
       }),
     );
-    expect(result.guests[0].paidConfirmed).toBe(true);
-    expect(result.guests[1].paidConfirmed).toBe(false);
+    expect(result.guests[0]!.paidConfirmed).toBe(true);
+    expect(result.guests[1]!.paidConfirmed).toBe(false);
   });
 
   it('converts ISO start_datetime to Date', () => {
--- a/frontend/src/api/eventMapper.test.ts
+++ b/frontend/src/api/eventMapper.test.ts
@@ -20,6 +20,27 @@ describe('mapEvent', () => {
     expect(result.startDatetime!.getUTCHours()).toBe(18);
   });
 
+  it('maps my_paid_confirmed', () => {
+    expect(mapEvent(wireEvent({ my_paid_confirmed: true })).myPaidConfirmed).toBe(true);
+  });
+
+  it('defaults myPaidConfirmed to false when absent', () => {
+    expect(mapEvent(wireEvent()).myPaidConfirmed).toBe(false);
+  });
+
+  it('maps paid_confirmed on guests', () => {
+    const result = mapEvent(
+      wireEvent({
+        guests: [
+          { user_id: 'u1', name: 'Paid Guest', status: 'attending', paid_confirmed: true },
+          { user_id: 'u2', name: 'Unpaid Guest', status: 'attending' },
+        ],
+      }),
+    );
+    expect(result.guests[0].paidConfirmed).toBe(true);
+    expect(result.guests[1].paidConfirmed).toBe(false);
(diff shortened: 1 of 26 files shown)


leahpeker:
Splitting into two stacked PRs to stay under the 400-line limit — see the follow-ups linked shortly.

PR #1272, opened 2026-08-04
fix: 1238 - poll finalize bypasses capacity, plus-ones, payment gate

### LINKED ISSUE #1238: Redesign event polls — finalize bypasses capacity, plus-ones, and payment gate
## Context

`EventPoll`/`PollOption`/`PollVote` landed 2026-04-02 (`feat: split datetime polls into dedicated EventPoll system`). RSVP capacity limits + waitlist landed a week later, 2026-04-10 (#259). The payment-confirmation gate (Issue 1045) landed even later. Poll finalize was never updated to know about either.

`finalize_event_poll` in `backend/community/_polls.py` seats every yes-voter directly:

```python
for user_id in yes_voter_ids:
    EventRSVP.objects.update_or_create(
        event=event,
        user_id=user_id,
        defaults={"status": RSVPStatus.ATTENDING, "cancelled_at": None},
    )
```

This bypasses `_apply_rsvp_in_transaction` — the shared write path every other RSVP flow (member, public, host-managed) goes through. As a result, poll finalize:

- **Ignores `max_attendees`** — can seat unlimited yes-voters as `ATTENDING` regardless of capacity, with no waitlist fallback.
- **Ignores `has_plus_one`** — a yes-voter's plus-one is dropped entirely.
- **Bypasses the payment-confirmation gate** (Issue 1045) — a yes-voter is seated as `ATTENDING` with no `paid_confirmed_at` stamp and no chance to confirm payment. The payment-gate work had to add a dedicated regression test (`TestPollFinalizeDoesNotBypassGate`) specifically because this path re-gates the guest on their very next RSVP write, rather than gating them at finalize time — a workaround, not a fix.

This isn't a one-off oversight; it's the third feature poll finalize has now shipped blind to. Worth stepping back and redesigning polls against the RSVP system as it exists today, rather than patching a fourth workaround around the same direct-write pattern next time something changes.

## What needs to happen

- Audit `backend/community/_polls.py` and `backend/community/models/poll.py` end-to-end against the current RSVP model (capacity/waitlist, payment gate, plus-ones, host-managed RSVPs).
- Decide whether poll finalize should route yes-voters through `_apply_rsvp_in_transaction` (or an equivalent shared path) instead of writing `EventRSVP` rows directly, so it inherits capacity/waitlist/payment-gate behavior automatically instead of needing bespoke handling per feature.
- Consider whether polls need their own capacity story (e.g. "more yes votes than spots" at finalize time) rather than silently overselling.

## Acceptance criteria

- [ ] TODO

leahpeker:
## Overview

`finalize_event_poll()` wrote `EventRSVP` rows directly instead of routing through `_apply_rsvp_in_transaction()`, the shared write path every other RSVP flow uses. This meant poll finalize bypassed capacity limits, dropped plus-ones, skipped the payment-confirmation gate, could resurrect cancelled RSVPs, and ignored `rsvp_enabled`.

Fixes by replacing the direct-write loop with calls through the shared path (extracted into `backend/community/_poll_finalize.py` to keep `_polls.py` under the line cap), so finalize now inherits capacity/waitlist, payment-gate, plus-one, and cancellation semantics automatically instead of needing bespoke handling per feature.

**Design decisions:**
- Seating order: earliest `PollVote.voted_at` first (FIFO, matching the existing waitlist order). Note `voted_at` is `auto_now`, so it reflects last vote change, not first vote — the only ordering signal the model currently records.
- Payment gate: unpaid yes-voters are seated `WAITLISTED` (not auto-seated) and notified to confirm payment, rather than seated unpaid with no notification.
- Plus-ones: preserved from an existing RSVP row if present; new rows default to no plus-one (polls carry no +1 concept).

**Behavior change to flag for review:** on a paid event, finalizing now leaves unpaid yes-voters `WAITLISTED` instead of `ATTENDING`. Hosts finalizing a paid poll will see a smaller attending count than before — this is the correct outcome, but will look like a regression if not expected.

## Test plan

- [x] Scoped test run: `backend/tests/test_poll_finalize_seating.py` + `backend/tests/test_rsvp_payment_gate_bypass.py` (16 tests, all passing)
- [ ] TODO — manual QA: finalize a poll on a capacity-limited event and confirm waitlist behavior

## Breaking changes

Behavioral: paid-event poll finalize now waitlists unpaid yes-voters instead of seating them unpaid (see note above). No migration needed — existing rows untouched.

Closes #1238

### DIFF
--- a/backend/community/_poll_finalize.py
+++ b/backend/community/_poll_finalize.py
@@ -1,0 +1,77 @@
+from notifications.service import create_waitlist_promoted_notifications
+
+from community._event_rsvps import _apply_rsvp_in_transaction
+from community._validation import Code, ValidationException
+from community.models import Event, EventRSVP, PollAvailability, PollOption, RSVPStatus
+
+
+def _yes_votes(winning_option: PollOption) -> list:
+    """Winning option's yes-votes, oldest first.
+
+    Ordering makes who gets a seat under capacity deterministic rather than
+    arbitrary row order. voted_at is auto_now, so this is really "least recently
+    changed their vote" — the only ordering signal PollVote records.
+    """
+    return list(
+        winning_option.votes.filter(availability=PollAvailability.YES)
+        .select_related("user")
+        .order_by("voted_at")
+    )
+
+
+def _seat_voter(event: Event, user, existing: EventRSVP | None) -> bool:
+    """Seat one yes-voter through the shared RSVP write path.
+
+    return(bool): True if they were waitlisted for non-payment on a gated event.
+    """
+    # Polls carry no +1 concept; only an existing row's standing +1 survives.
+    has_plus_one = existing is not None and existing.has_plus_one
+    try:
+        _apply_rsvp_in_transaction(event.id, user, RSVPStatus.ATTENDING, has_plus_one)
+        return False
+    except ValidationException as exc:
+        if exc.code != Code.Event.PAYMENT_CONFIRMATION_REQUIRED:
+            raise
+    # Unpaid on a gated event: waitlist rather than seat unconfirmed or drop them,
+    # so confirming payment promotes them through the normal path.
+    _apply_rsvp_in_transaction(event.id, user, RSVPStatus.WAITLISTED, has_plus_one)
+    return True
+
+
+def _seat_or_skip(event: Event, user, existing: EventRSVP | None) -> bool:
+    """Seat a voter, absorbing the reasons a single voter can't be seated.
+
+    return(bool): True if they were waitlisted for non-payment on a gated event.
+    """
+    # A member who explicitly said they can't go must not be resurrected by finalize.
+    if existing is not None and existing.status == RSVPStatus.CANT_GO:
+        return False
+    try:
+        return _seat_voter(event, user, existing)
+    except ValidationException:
+        # RSVPs disabled/closed for this event, or this voter can no longer see
+        # it — one unseatable voter must not fail the whole finalize.
+        return False
+
+
+def seat_yes_voters(event: Event, winning_option: PollOption) -> None:
+    """Seat the winning option's yes-voters via the shared RSVP write path.
+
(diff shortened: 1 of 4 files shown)


leahpeker:
### Code review

Found 1 issue:

1. Unpaid poll voters are notified as "waitlist promoted" when they were never promoted

`seat_yes_voters` collects voters who got waitlisted because they're unpaid on a payment-gated event, then calls `create_waitlist_promoted_notifications(event, unpaid_user_ids, unpaid_user_ids)` — passing the same list as both `promoted_user_ids` and `unpaid_user_ids`.

https://github.com/ProteinDeficientsAnonymous/pda/blob/a58f43d69ef67f6f00a96424f0d97ac60b5e52d6/backend/community/_poll_finalize.py#L71-L77

`create_waitlist_promoted_notifications` stamps every id in `promoted_user_ids` with `notification_type=NotificationType.WAITLIST_PROMOTED`:

https://github.com/ProteinDeficientsAnonymous/pda/blob/a58f43d69ef67f6f00a96424f0d97ac60b5e52d6/backend/notifications/service.py#L231-L257

These voters went straight from no RSVP to `WAITLISTED` via the payment gate — they were never previously seated or waitlisted, so there's nothing to be "promoted" from. Compare the legitimate caller in `promote_from_waitlist`, where `promoted_user_ids` are users whose status was just changed to `ATTENDING` and `unpaid_user_ids` is a genuinely smaller subset:

https://github.com/ProteinDeficientsAnonymous/pda/blob/a58f43d69ef67f6f00a96424f0d97ac60b5e52d6/backend/community/_event_helpers.py#L184-L190

The message copy happens to read correctly ("your spot isn't confirmed until you pay") since `user_id in unpaid` is trivially true here, but the persisted `notification_type` is wrong on every payment-gated poll finalize that leaves a yes-voter unpaid.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1308, opened 2026-08-06
feat(join): add configurable question types and shared authoring UI

timkwist:
Move question rendering into shared components, add join textarea/select
authoring, and thin-wrap survey/join dialogs over QuestionAuthorDialog.

Co-authored-by: Cursor <cursoragent@cursor.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


backend/tests/test_join_form_admin.py
```diff
@@ -77,3 +96,43 @@ def test_reorder_unauthenticated(self, api_client):
...
+                    "required": True,
+                }
+            ),
+            content_type="application/json",
+            **form_admin_headers,
+        )
+        assert response.status_code == 201
+        body = response.json()
+        assert body["field_type"] == "textarea"
+        assert JoinFormQuestion.objects.get(id=body["id"]).field_type == "textarea"
```
leahpeker: nit: can use enum instead of raw "textarea" text

frontend/src/components/questions/QuestionAuthorDialog.test.tsx
```diff
@@ -0,0 +1,60 @@
+import { render, screen } from '@testing-library/react';
```
leahpeker, line 1: could we call this like eventQuestionnaire instead of questions? just to make it clearer

and then like QuestionnaireField QuestionnaireAuthor?

frontend/src/components/questions/QuestionAuthorDialog.tsx
```diff
@@ -0,0 +1,150 @@
+import type { SyntheticEvent } from 'react';
```
leahpeker, line 1: i'm confused a lil on the naming of this too actually -- is this like a CreateQuestionnaireModal?

frontend/src/components/questions/QuestionAuthorDialog.test.tsx
leahpeker: or wait these are shared, so they dont belong to the questionnaire then? my b

frontend/src/components/questions/QuestionAuthorDialog.tsx
leahpeker: actually i see it's not -- could you describe what this one is? and is dialog a modal?

frontend/src/components/questions/questionTypeOptions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export interface QuestionTypeOption {
+  value: QuestionType;
+  label: string;
+  wantsOptions: boolean;
+}
+
+/** Canonical authoring metadata for every catalog question type. */
+export const QUESTION_TYPE_OPTION_BY_TYPE = {
+  text: { value: 'text', label: 'short text', wantsOptions: false },
```
leahpeker: nit: use enums instead of the raw text and also instead of typing it out like this maybe we can use a map to map the values of the question types to the value, label, wants options struct

frontend/src/screens/admin/joinQuestionTypeOptions.ts
```diff
@@ -0,0 +1,13 @@
...
+  type QuestionTypeOption,
+} from '@/components/questions/questionTypeOptions';
+
+const JOIN_QUESTION_TYPE_OPTION_BY_TYPE = {
+  text: QUESTION_TYPE_OPTION_BY_TYPE.text,
+  textarea: QUESTION_TYPE_OPTION_BY_TYPE.textarea,
+  select: QUESTION_TYPE_OPTION_BY_TYPE.select,
+} satisfies Record<JoinQuestionType, QuestionTypeOption>;
+
+export const JOIN_QUESTION_TYPE_OPTIONS = Object.values(JOIN_QUESTION_TYPE_OPTION_BY_TYPE);
```
leahpeker: there's soemthing really similar in another file -- i think maybe this is duplicated

frontend/src/screens/admin/joinQuestionTypeOptions.ts
```diff
@@ -0,0 +1,13 @@
...
+import {
+  QUESTION_TYPE_OPTION_BY_TYPE,
+  type QuestionTypeOption,
+} from '@/components/questions/questionTypeOptions';
+
+const JOIN_QUESTION_TYPE_OPTION_BY_TYPE = {
+  text: QUESTION_TYPE_OPTION_BY_TYPE.text,
+  textarea: QUESTION_TYPE_OPTION_BY_TYPE.textarea,
+  select: QUESTION_TYPE_OPTION_BY_TYPE.select,
+} satisfies Record<JoinQuestionType, QuestionTypeOption>;
```
leahpeker: never seen satisfies used before lol

frontend/src/screens/surveys/SurveyScreen.tsx
```diff
@@ -114,7 +113,7 @@ function SurveyForm({ survey }: { survey: Survey }) {
           </Button>
         ) : null}
 
-        {submit.isSuccess && !readOnly ? <p className="text-success text-sm">saved ✓</p> : null}
+        {submit.isSuccess && !readOnly ? <p className="text-sm text-green-700">saved ✓</p> : null}
```
leahpeker: go back to success -- that is for dark mode

leahpeker (approved):

frontend/src/components/questions/QuestionAuthorDialog.test.tsx
timkwist: Yep — leaving the shared `questions/` names. `QuestionAuthorDialog` / `QuestionField` are reused by survey + join authoring, so a questionnaire-specific name would be misleading.

frontend/src/components/questions/QuestionAuthorDialog.tsx
timkwist: Shared create/edit dialog for a single catalog question (label, type, options, required). Survey and join each thin-wrap it with their type subset.

And yes — in this codebase `Dialog` is the modal primitive (`@/components/ui/Dialog`).

frontend/src/screens/admin/joinQuestionTypeOptions.ts
timkwist: `satisfies` checks the object against the type without widening the inferred type — handy for keyed maps like this. Kept it on the join subset list via `as const satisfies readonly JoinQuestionType[]`.

PR #1314, opened 2026-08-06
feat(events): add host RSVP response reporting

timkwist:
Expose saved answer snapshots only to event managers and keep response
reports available after questions, RSVPs, or the event itself close.

Co-authored-by: Cursor <cursoragent@cursor.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/frontend/src/screens/events/EventDetailKebabMenu.tsx
+++ b/frontend/src/screens/events/EventDetailKebabMenu.tsx
@@ -10,6 +10,7 @@ import { Feature } from '@/models/featureFlags';
 import { EmailBlastDialog } from './EmailBlastDialog';
 import { canEditEvent } from './eventEditGate';
 import { GroupTextDialog } from './GroupTextDialog';
+import { hasSavedRsvpAnswers } from './rsvpQuestions';
 
 interface Props {
   event: Event;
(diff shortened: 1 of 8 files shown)


frontend/src/screens/events/EventDetailKebabMenu.tsx
```diff
@@ -84,7 +87,7 @@ export function EventDetailKebabMenu({ event, eventHasEnded, canManageRsvps }: P
                 setOpen(false);
               }}
             >
-              manage rsvps
+              {eventHasEnded ? 'question responses' : 'manage rsvps'}
```
timkwist: if we had it as manage rsvps even after an event ended in the past, we should probably just keep it that way right?

frontend/src/screens/events/EventRsvpResponsesSection.tsx
```diff
@@ -0,0 +1,159 @@
...
+        ))}
+      </ul>
+    </div>
+  );
+}
+
+function statusLabel(status: string): string {
+  if (status === RsvpServerStatus.Attending) return 'going';
+  if (status === RsvpServerStatus.Maybe) return 'maybe';
+  if (status === RsvpServerStatus.Waitlisted) return 'waitlist';
```
timkwist: magic strings that maybe we could constant? maybe / going / waitlist

frontend/src/screens/events/EventRsvpResponsesSection.tsx
```diff
@@ -0,0 +1,159 @@
...
+  return status;
+}
+
+function renderAnswerForColumn(
+  snap: { label: string; answer: string } | undefined,
+  column: ResponseColumn,
+): string {
+  if (snap?.label !== column.label) return '—';
+  return snap.answer.trim().replaceAll(',', ', ') || '—';
```
timkwist: why return '-' as default?

leahpeker (approved):

PR #1336, opened 2026-08-08
fix: 1332 - check-in report counts explicit no-show marks

### LINKED ISSUE #1332: check in report bug
```
there's still a person shown as unmarked in the check in report but this person is marked as didnt go

their original rsvp was maybe
```

## Metadata

- **Route:** `/events/052e66e7-9f2f-4c19-af76-99894b576f20/report`
- **User Agent:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36`

leahpeker:
## Overview

Original report: a guest showed as "unmarked" in the event check-in report even though a host explicitly marked them "didn't attend" — their original RSVP had been "maybe".

**Scope evolved during review** (see PR discussion): no-show is fixed repo-wide as "RSVP'd going (`status == ATTENDING`) and marked `DIDNT_GO`" — see the shared `is_no_show` helper in `backend/community/_rsvp_counts.py`, used consistently by the check-in report, the admin attendance report, and event stats. Broadening the no-show bucket itself wasn't the right fix.

Instead, the check-in report gets a distinct **didn't go** bucket: anyone marked `DIDNT_GO` who isn't a true no-show (wasn't RSVP'd `ATTENDING`) and didn't originally say they couldn't go (not `CANT_GO`) now lands there instead of the catch-all "unmarked" — this is the actual fix for the original bug.

Also includes a smaller, separate fix found along the way: `_build_report` checked the no-show condition before `CANT_GO`, so a guest who told us in advance they weren't coming (with a stray no-show mark on their RSVP) could get miscategorized as a no-show instead of staying "canceled". Reordered so `CANT_GO` is checked first.

## Test plan

- Full `backend/tests/test_event_check_in_report.py` suite passes (25 passed).
- New `test_maybe_rsvp_marked_didnt_attend_is_in_didnt_go_not_unmarked` covers the original bug scenario directly.

Closes #1332

### DIFF
--- a/backend/community/_event_report.py
+++ b/backend/community/_event_report.py
@@ -60,10 +60,6 @@ def _build_report(event: Event, viewer) -> CheckInReportOut:
     attended, no_shows, canceled, unmarked = [], [], [], []
     for rsvp in _report_rsvps(event):
         base = _person(rsvp, viewer, can_see_phones)
-        # ATTENDED overrides status (a host can check in a "maybe"); a no-show
-        # only counts if they were actually RSVP'd ATTENDING. CANT_GO is
-        # checked before no-show so a guest who told us in advance they
-        # weren't coming stays "canceled" even if a stray no-show mark exists.
         if rsvp.attendance == AttendanceStatus.ATTENDED:
             attended.append(
                 AttendedPersonOut(**base.model_dump(), checked_in_at=rsvp.checked_in_at)
--- a/backend/community/_event_report.py
+++ b/backend/community/_event_report.py
@@ -17,10 +17,10 @@ from community._event_report_schemas import (
     CheckInReportPersonOut,
 )
 from community._events import _can_edit_event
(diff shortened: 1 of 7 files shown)


leahpeker:
Follow-up fix: the previous commit removed the `ATTENDING` gate on the no-show check but left `CANT_GO` after it in the if/elif chain — so a cant-go RSVP with a stray no-show mark would've wrongly bucketed as "no-show" instead of "canceled". Reordered so `CANT_GO` is checked before `NO_SHOW`. Caught in review before merge — thanks for the sharp eye.

`test_cant_go_rsvp_marked_didnt_attend_is_not_a_no_show` is back to its original assertion; the maybe-RSVP no-show test (the actual Issue 1332 bug) still passes.

